PR #15, opened 2018-09-19
Sl generation

ttgenius:
1.seperate write from generate sk
2.read from header and context template
3. More vars in sl template

### DIFF
--- a/scripts/checkpoint.py
+++ b/scripts/checkpoint.py
@@ -1,0 +1,50 @@
+import os
+import glob
+
+# Examples:
+# sim_waveform_dirs =
+# ['/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434',
+# '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP29-29_S1524',
+# '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP07-29_S1304']
+# dire = '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434'
+# output_sim_dir = /nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434/../../../IM_calc/Kelly_HYP20-29_S1434/
+
+
+def checkpoint_sim_obs(waveform_dirs, relative_path_to_im_calc):
+    """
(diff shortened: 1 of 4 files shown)


scripts/checkpoint.py
```diff
@@ -0,0 +1,50 @@
...
+# Examples:
+# sim_waveform_dirs =
+# ['/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434',
+# '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP29-29_S1524',
+# '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP07-29_S1304']
+# dire = '/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434'
+# output_sim_dir = /nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly/BB/Cant1D_v3-midQ_OneRay_hfnp2mm+_rvf0p8_sd50_k0p045/Kelly_HYP20-29_S1434/../../../IM_calc/Kelly_HYP20-29_S1434/
+
+
+def checkpoint_sim_obs(waveform_dirs, relative_path_to_im_calc):
```
shot0829: Dont use relative path. 
you have the run_dir, sim_dir. fault_name, generate the full-path and enforce the rule.
example pseudo code: 
`IM_calc_dir = join(sim_dir,'IM_calc');
for hyp in hpy_list: 
do
     IM_hpy = join(IM_calc_dir,hyp);
`

scripts/checkpoint.py
```diff
@@ -0,0 +1,50 @@
...
+    :return: a list of not completed waveform dirs
+    """
+    for directory in waveform_dirs[:]:
+        dir_name = directory.split('/')[-1]
+        if dir_name == 'IM_calc':
+            waveform_dirs.remove(directory)
+        else:
+            output_dir = os.path.join(directory, relative_path_to_im_calc, dir_name)
+            if os.path.isdir(output_dir):  # if output dir exists
+                sum_csv = glob.glob1(output_dir, '*.csv')
```
shot0829: make a test function that takes sim_dir, i.e "/nesi/nobackup/nesi00213/RunFolder/Cybershake/v18p6/Runs/test/Kelly" and Srf name ( Kelly_HYP01-02_S1244).
returns failed or success.
This will allow user/scripts to call the test function if only going to test for one single realization/srf under a simulation.

shot0829: the function hierarchy should be something like this:
test_for_all_runs
----test_for_sim
---------test_for_a_realization

scripts/checkpoint.py
ttgenius: if no using relative path, function would be:
def checkpoint_sim_obs(run_dir, waveform_dirs):
    for directory in waveform_dirs[:]:
        dir_name = directory.split('/')[-1]
        if dir_name == 'IM_calc':
            waveform_dirs.remove(directory)
        else:
           fault_name = dir_name.split('_')[0]
            output_dir = os.path.join(run_dir, fault_name, 'IM_calc')
            if os.path.isdir(output_dir):  # if output dir exists
                sum_csv = glob.glob1(output_dir, '*.csv')
                meta = glob.glob1(output_dir, '*imcalc.info')
                # if sum_csv and meta are not empty lists('.csv' and '_imcalc.info' files present)
                # then we think im calc on the corresponding dir is completed and hence remove
                if sum_csv and meta:
                    waveform_dirs.remove(directory)
    return waveform_dirs

PR #23, opened 2018-10-05
Uses a different random number seed for a different rank

sungeunbae:
This is to avoid all processes using the same random seed and effectively producing similar random patterns for stations. 

If we have more processes than a number of stations, each process computes one station (or no station), and as a result, all stations showed very similar result due to using the same random sequence. 

This fix can mitigate the problem by adding "rank" to the "seed", forcing every rank to produce different random sequence.


jasonmotha:
@viktor76525 Would you like to have a look at this quick change?

PR #32, opened 2018-12-10
Params

ttgenius:
1.Must run with yaml branch of qcore
sample cmd:
```python ~/slurm_gm_workflow/scripts/install.py --user_root ~/Albury_newman/ --srf_dir ~/Albury_newman/Data/Sources/ --vm_dir ~/Albury_newman/Data/VMs/ --version gmsim_v18.5.3```

2.change 'bin_process_ver' in install.py and 'templates_dir' in workflow_config.json if necessary
3.only work with manual workflow
4.Must cp a copy of machine.env into your simulation dir. eg
```
source /home/melody.zhu/.bashrc
export BINPROCESS=/home/melody.zhu/slurm_gm_workflow/scripts
```

5.Note the strict name convention of emod3d_defaults_gmsim_v18.5,3.yaml, where gmsim_v18.5.3 must match the version of the simulation exactly. And this yaml should be present in the templates_dir
5.tested with Albury benchmark with 'main_local'
6.some comments or unused code lines are left as is
7.added test code for manual install

### CODE REVIEW

File: install/create_config_file.py [hunk 1]
```diff
@@ -10,7 +10,7 @@
 bin_process_path = os.path.join(global_root, "workflow/scripts")
 qcore_path =  os.path.join(global_root, "qcore")
 tools_path = os.path.join(global_root, "tools")
-
+runfolder_path = "/home/melody.zhu/project/nesi00213/RunFolder"
```
claudio525: hardcoded user specific path in a config?

scripts/install.py
```diff
@@ -252,163 +255,116 @@ def action(sim_dir, event_name, run_name, run_dir, vel_mod_dir, srf_dir, srf_sto
...
+    fault_params_dict = {}
+    sim_params_dict = {}
+    vm_params_dict = {}
+
+    sim_params_dict['run_name'] = run_name
+
+    # select during install
+    root_params_dict['version'] = version
+
+    root_params_dict['dt'] = 0.02
```
claudio525: shouldn't this be the dt parameter passed into the function?

scripts/install.py
```diff
@@ -228,14 +230,15 @@ def q_final_confirm(run_name, yes_statcords, yes_model_params):
...
-def action(sim_dir, event_name, run_name, run_dir, vel_mod_dir, srf_dir, srf_stoch_pairs, params_vel_path,
-           stat_file_path, vs30_file_path, vs30ref_file_path, MODEL_LAT, MODEL_LON, MODEL_ROT, hh, nx, ny, nz, sufx,
-           sim_duration, flo, vel_mod_params_dir, yes_statcords, yes_model_params,dt = default_dt, hf_dt = default_hf_dt):
-    lf_sim_root_dir, hf_dir, bb_dir, figures_dir = os.path.join(sim_dir, "LF"), os.path.join(sim_dir,
-                                                                                             "HF"), os.path.join(
-        sim_dir, "BB"), os.path.join(sim_dir, "Figures")
+def create_params_dict(version, sim_dir, event_name, run_name, run_dir, vel_mod_dir, srf_dir, srf_file, stoch_file,
+                       params_vel_path, stat_file_path, vs30_file_path, vs30ref_file_path, MODEL_LAT, MODEL_LON,
+                       MODEL_ROT, hh, nx, ny, nz, sufx, sim_duration, flo, vel_mod_params_dir, yes_statcords,
+                       yes_model_params, dt=default_dt, hf_dt=default_hf_dt):
```
claudio525: Parameters flo, hf_dt and srf_file don't appear to be used in the function?

scripts/workflow_config.json
```diff
@@ -3,11 +3,10 @@
   "emod3d_version" : "3.0.4",
   "global_root" : "/nesi/project/nesi00213",
   "bin_process_path" : "/nesi/project/nesi00213/opt/maui/emod3d/3.0.4-gcc/bin/",
-  "qcore_lib_path" : "/nesi/project/nesi00213/qcore",
+  "qcore_lib_path" : "/home/melody.zhu/qcore",
```
claudio525: user directory path?

scripts/install.py
ttgenius: good spot

install/create_config_file.py
ttgenius: all changed back to global

scripts/install.py
ttgenius: good spot. flo and hf_dt now in dummpy template; srf_file is used.

testing/test_manual_install/test_manual_install.py
```diff
@@ -52,7 +98,8 @@ def test_e3ds(bench_path, test_path):
         out, err = shared.exe(
             'diff {} {}'.format(os.path.join(bench_path, bench_e3ds[i]), os.path.join(test_path, test_e3ds[i])))
         logs += out + err
-    assert logs == ''
+
+# assert logs == ''
```
claudio525: Remove dead code?

testing/test_manual_install/test_manual_install.py
```diff
@@ -1,16 +1,47 @@
...
+import shutil
 from qcore import timeseries
 from qcore import shared
+from qcore import utils
 
 
-def check_values(s1, s2):
+DIVIDER = '-' *20
+TXT1 = 'txt1'
+TXT2 = 'txt2'
```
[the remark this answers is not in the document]
ttgenius: All changed according to per comments! Thanks!

scripts/install.py
```diff
@@ -252,163 +258,112 @@ def action(sim_dir, event_name, run_name, run_dir, vel_mod_dir, srf_dir, srf_sto
...
-    #    print "Done"
-    #    with open(os.path.join(sim_dir,"params_base.py"),"a") as f:
-    #        f.write("stat_file='%s'\n" %stat_file_path)
-    #        f.write("STAT_FILES=[stat_file]\n")
-    #
-    #    print "Statgrid: %s" %stat_file_path
-
-
-
-    show_horizontal_line(c='*')
```
shot0829: i think it is better to keep this line, so the user can tell what is going on more easily

scripts/install_bb.py
```diff
@@ -32,224 +31,182 @@ def q0():
...
             print"       If you need to specific hf_kappa and hf_sdrop value for each SRF, add hf_kappa_list and hf_sdrop_list to params_base.py"
 
-    else:
-        print "hf_kappa_list: ", hf_kappa_list
-        print "hf_sdrop_list: ", hf_sdrop_list
-        print "srf_files:", srf_files
-        if len(hf_kappa_list) != len(hf_sdrop_list) or len(hf_kappa_list) != len(srf_files):
-            print "Error: hf_kappa_list (len=%d), hf_sdrop_list (len=%d) and srf_files (len=%d) should be of the same length."%(len(hf_kappa_list),len(hf_sdrop_list), len(srf_files))
-            sys.exit()
+    #dunno what the following does, old_prams always=True
```
shot0829: You should be more responsible and ask around for answers before you made drastic changes.
The 'old_params' was there because at one point we decided to support both 'single' and 'multiple' SRF per simulation. 
But now since we are enforcing single SRF only, any related script shuold be removed.

ttgenius:
It is guaranteed that only one srf will be stored as a string in the yaml
file(happened at install.py) So probably no need to check if there's one
srf inside the params.srf_file string?
eg.
params.srf_file="/home/melody.zhu/Albury_newman/Data/Sources/Albury/Srf/Albury_HYP15-21_S1384.srf"

PR #39, opened 2019-01-13
Qsw 648 estimate wall clock

claudio525:
Added wall clock estimation model and updated the submit scripts.
Lots of changes to make the workflow python3 compatible.

Depends on PR https://github.com/ucgmsim/qcore/pull/42


metadata/agg_json_data.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if len(cur_columns) > len(columns):
+            columns = cur_columns
+
+    if len(columns) > 0:
+        row_data = get_row_data(data_dict, columns)
+
+        # Remove empty rows
+        empty_rows_mask = np.all(pd.isnull(row_data), axis=1)
+        row_data = row_data[~empty_rows_mask]
```
sungeunbae: clever :)

metadata/write_jsons.py
```diff
@@ -0,0 +1,256 @@
...
+                    if key == 'run_time':
+                        data[i] = "{:.5f} hour".format(float(data[i]))
+                    result_dict[realization][sim_type][key] = data[i]
+
+                # Extra LF metadata
+                if sim_type == 'LF' and \
+                        result_dict[realization][sim_type].get('total_memory_usage') is None:
+                    # Get the rlog dir for the current realization
+                    cur_rlog_dir = [rlog_dir for rlog_dir in rlog_dirs if
+                                    realization in rlog_dir.split("/")][0]
```
sungeunbae: Any chance that this list is empty, and runs into IndexError?

metadata/write_jsons.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    :param realization_rlog_dir: **/LF/**/Rlog
+    :return:total_memory usage for each realization in GB
+    """
+    os.chdir(realization_rlog_dir)
+    cmd = "grep 'total for model' *.rlog"
+    output = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True).communicate()[0].decode()
+    total = 0
+    unit = ''
+
```
sungeunbae: I sometimes find it helpful to show output example before playing with strip() and split() - it makes the future debugging extremely easy (when *.rlog changes its format and the output suddenly looks different, we can immediately know what's going on). You don't need to change anything - just my opinion...

metadata/write_jsons.py
```diff
@@ -0,0 +1,256 @@
...
+    :param sim_type: [BB|HF|LF]
+    :param realization: eg.Hossack_HYP222-501_S3454
+    :return: submit_time
+    """
+    if sim_type == 'LF':
+        sim_type = 'EMOD3D'
+    time_file = os.path.join(ch_log_dir, "{}.{}.log".format(sim_type, realization))
+    if os.path.isfile(time_file):
+        with open(time_file, 'r') as f:
+            submit_time = f.read().split(':')[-1].strip()  # 20181031_060458
```
sungeunbae: Aha, you do give examples for harder ones. Good :)

metadata/write_jsons.py
```diff
@@ -0,0 +1,256 @@
...
+
+def write_fault_jsons(fault_dir, single_json):
+    """
+    write json file(s) for a single fault
+    :param fault_dir:
+    :param single_json: boolean, user input
+    :return:
+    """
+    all_sims_dict = get_all_sims_dict(fault_dir)
+    if all_sims_dict != None:
```
sungeunbae: is not None?

scripts/set_runparams.py
```diff
@@ -82,5 +77,11 @@ def create_run_params(sim_dir, srf_name=None):
 
 
 if __name__ == '__main__':
+    workflow_config = load_config.load(
+        os.path.dirname(os.path.realpath(__file__)), "workflow_config.json")
+    global_root = workflow_config["global_root"]
+    tools_dir = os.path.join(global_root, 'opt/maui/emod3d/3.0.4-gcc/bin')
```
sungeunbae: Should this be hard-coded?

scripts/submit_emod3d.py
```diff
@@ -1,196 +1,138 @@
...
+    """Populates the template and writes the resulting slurm script to file"""
+
+    with open('run_emod3d.sl.template', 'r') as f:
+        template = f.read()
+
+    replace_t = [("{{lf_sim_dir}}", lf_sim_dir), ("{{tools_dir}}", tools_dir),
+                 ("{{mgmt_db_location}}", mgmt_db_location),
+                 ("{{sim_dir}}", sim_dir), ("{{srf_name}}", srf_name)]
+
+    for pattern, value in replace_t:
```
sungeunbae: 👍

scripts/submit_emod3d.py
```diff
@@ -1,196 +1,138 @@
...
-        elif args.set_params_only == True:
-            submit_yes = False
-        else:
-            submit_yes = confirm("Also submit the job for you?")
-        print("params.srf_file", params.srf_file)
-            #get the srf(rup) name without extensions
+    workflow_config = load_config.load(
+        os.path.dirname(os.path.realpath(__file__)), "workflow_config.json")
+    global_root = workflow_config["global_root"]
+    tools_dir = os.path.join(global_root, 'opt/maui/emod3d/3.0.4-gcc/bin')
```
sungeunbae: Same as one of earlier comments. Should this be hard-coded?

shared_workflow/load_config.py
```diff
@@ -1,22 +1,18 @@
-
-import sys
 import os
 import json
 
-def load(directory=os.path.dirname(os.path.abspath(__file__)), cfg_name="workflow_config.json"):
-    #directory = os.path.dirname(os.path.abspath(__file__))
-    print directory
-    #config_file = os.path.join(directory, "workflow_config.json")
+# Shouldn't this be in shared or qcore?
```
sungeunbae: Fair point. Who is using this?

metadata/write_jsons.py
claudio525: Yeah, good catch. Will fix.

metadata/write_jsons.py
claudio525: I didn't actually write that particular function, I think the reason why github shows the whole file as new is because I moved the file, which is a bit annoying. I did modify this file heavily as well though. 

Yeah I agree, I also don't think this function should be using an external cmd, this should be done in python in my opinion. I have been modifying this script quite heavily recently, so I'll change that next time I'm working on it (the most up to date version of this file is in a different branch).

PR #43, opened 2019-01-18
IM calc estimation + format

claudio525:


templates/im_calc.sl.template
```diff
@@ -34,13 +32,16 @@ echo ___calculating simulations___
         fi
         #TODO: adjust this temporary code to use the update_db once merged
         cd {{mgmt_db_location}}/Runs/{{fault_name}}/{{sim_name}}
-        fd_name=`python -c "from qcore import utils; p = utils.load_yaml('sim_params.yaml'); print p['FD_STATLIST']"`
+        fd_name=`python -c "from qcore import utils; params = utils.load_sim_params('{{sim_dir}}/../sim_params.yaml'); print(params.FD_STATLIST)"`
```
jasonmotha: Is there a bash way to do this that is slightly more elegant than calling a python command like this? Even if we create our own scriptlet to return the value of a yaml file it would be nicer.

templates/im_calc.sl.template
```diff
@@ -34,13 +32,16 @@ echo ___calculating simulations___
...
         fd_count=`cat $fd_name | wc -l`
-        nt=`python -c "from qcore import utils; d = utils.load_sim_params('sim_params.yaml', load_vm=False); print int(float(d.sim_duration) / float(d.hf.hf_dt))"`
+        nt=`python -c "from qcore import utils; params = utils.load_sim_params('{{sim_dir}}/../sim_params.yaml'); nt = float(params.sim_duration)/float(params.hf.hf_dt); print(int(nt))"`
         
         run_time=`python -c "from datetime import datetime; tdelta= datetime.strptime('$end_time', '$runtime_fmt') - datetime.strptime('$start_time', '$runtime_fmt'); print tdelta.total_seconds()/3600"`
 
-        echo "{{sim_name}} IM_calc $SLURM_JOBID {{np}} $fd_count $nt $run_time $start_time $end_time" >> {{mgmt_db_location}}/Runs/{{fault_name}}/{{sim_name}}/ch_log/IM_calc.$timestamp.$SLURM_JOBID.log
+        # Get the number of pSA IMs and components out of the result csv file
+        psa_count=`python -c "with open('{{sim_dir}}/../IM_calc/{{sim_name}}/{{sim_name}}.csv', 'r') as f: print(sum([1 for header in f.readline().rstrip().split(',') if 'pSA' in header]))"`
```
jasonmotha: Similar, I feel like using a bash swiss army knife is more appropriate than python. Like even piping the first line into python would be better than reading the file in python.

PR #47, opened 2019-01-23
Qsw 669 - cybershake estimation

claudio525:
Also adds number of cores scaling to LF again.


metadata/agg_json_data.py
```diff
@@ -142,7 +142,7 @@ def clean_df(df):
                     )
                 # Count the number of components calculated
                 elif meta_col == MetaConst.im_comp.value:
-                    df.loc[:, (sim_type, meta_col)] = [
+                    df.loc[:, (sim_type, MetaConst.im_comp_count.value)] = [
```
sungeunbae: I guess this was a bug?

estimation/estimate_wct.py
```diff
@@ -270,7 +416,7 @@ def estimate(
     data = scaler.transform(input_data)
     wc = model.predict(data)
 
-    return wc
+    return wc.reshape(-1)
```
sungeunbae: Just a question - what is the original shape of wc supposed to be? So previously you needed [0][0] to get core_hours, and reshape[-1] removed that requirement?

metadata/agg_json_data.py
claudio525: Yeah this was a bug. It meant that the component count was stored in the column that initially contained the component string (i.e. "['geom',..]") instead of in a separate column.

estimation/estimate_wct.py
claudio525: So the [0][0] is from the functions that only return a single estimation and I think when I merged master back into this branch (which contained some other changes to the single estimation functions) the [0][0] ended up on the functions that do bulk estimation, which doesn't work. 
I did some quick debugging to see what the keras model actually returns, which is array of shape (-1, 1), so I just reshape it to be 1-D and everything works nicely from there.

PR #51, opened 2019-02-04
Ps perturbation

jpa198:


scripts/cybershake/install_cybershake.py
```diff
@@ -22,8 +22,12 @@ def main():
     parser.add_argument("--n_rel", type=int, default=None, help="the number of realisations expected")
 
     args = parser.parse_args()
-    
-    cybershake_cfg = ldcfg.load(os.path.dirname(args.config), os.path.basename(args.config))
+    try:
+        # try to read the config file
+        cybershake_cfg = ldcfg.load(os.path.dirname(args.config), os.path.basename(args.config))
+    except:
+        print("Error while parsing the config file, please double check inputs.")
```
jasonmotha: Best to print the exception too to give us a bit more information as to why it gave up.

PR #52, opened 2019-02-05
Bb vs30ref

jasonmotha:


scripts/bb_sim.py
```diff
@@ -105,7 +106,7 @@
             )
         ).get
     )(lf.stations.name)
-    assert not np.isnan(vsites).any()
+    assert not np.isnan(vs30s).any()
```
claudio525: What about using an if block instead of a try/except?

PR #56, opened 2019-02-08
Machine var

jpa198:


shared_workflow/shared.py
```diff
@@ -110,6 +111,46 @@ def get_vs(source_file):
...
+
+    for (template_sig, replacement) in replacements:
+        full_text = full_text.replace(template_sig, replacement)
+
+    return full_text
+
+
+def get_partition(machine, core_hours=None):
+    if machine == 'maui':
+        part = "nesi_research"
```
jasonmotha: Should use partition for clarity.

PR #64, opened 2019-02-15
Qsw 696 - Progress

claudio525:
- Adds the progress script.
- Changed cybershake estimation to use constants.
- The diff shows dashboard.py as a new file,  but this was already merged in PR https://github.com/ucgmsim/slurm_gm_workflow/pull/60 , so I think thats just a github issue....

Requires PR https://github.com/ucgmsim/qcore/pull/72


dashboard/cybershake_progress.py
```diff
@@ -0,0 +1,373 @@
...
+    with open(cybershake_list, "r") as f:
+        lines = f.readlines()
+
+    fault_names, r_counts = [], []
+    for ix, line in enumerate(lines):
+        line_l = line.split(" ")
+        fault_names.append(line_l[0].strip())
+
+        try:
+            r_counts.append(int(line_l[1].strip().rstrip("r")))
```
jasonmotha: We need to standardise the 'r'

dashboard/cybershake_progress.py
```diff
@@ -0,0 +1,373 @@
...
+                    progress_df.loc[fault, ("total", COMPLETED_R_COUNT_COL)],
+                    progress_df.loc[fault, ("total", R_COUNT_COL)],
+                ),
+                "True"
+                if progress_df.loc[fault, ("total", MISSING_DATA_FLAG_COL)]
+                else "False",
+            )
+        )
+
+    lf_act = progress_df[const.ProcessType.EMOD3D.str_value, ACT_CORE_HOURS_COL].sum()
```
jasonmotha: It almost seems nicer if we loop it. But that has disadvantages too.

dashboard/cybershake_progress.py
jasonmotha: https://github.com/ucgmsim/slurm_gm_workflow/blob/master/scripts/cybershake/install_cybershake.sh#L28

NKR are all currently valid realisation specifications. A conversation about wether we are deprecating the old ones is a worthwhile.

dashboard/cybershake_progress.py
claudio525: Good catch, no idea why I did it that way, its rather ugly....

dashboard/cybershake_progress.py
claudio525: Have added "n" and "k"

PR #67, opened 2019-02-18
Qsw 682

claudio525:
- Some tidy up of install
     - Moved shared install functions into install_shared.py
- Moved fd files into fault level for cybershake install
- Moved fd file params into fault yaml params instead of sim


scripts/cybershake/install_cybershake_fault.py
```diff
@@ -54,72 +62,66 @@ def main():
...
     if not valid_vm:
-        message = "Error: VM {} failed {}".format(source, message)
+        message = "Error: VM {} failed {}".format(fault_name, message)
         print(message)
         with open(error_log, "a") as error_fp:
             error_fp.write(message)
         exit()
 
+    # Load the variables from params_vel.py
     params_vel_path = os.path.join(vel_mod_dir, params_vel)
```
jasonmotha: So we end up with a params vel.py, params_vel.json and vm_params.yaml?

claudio525: I believe this is the current state of things. I agree its seem a bit over the top haha.

PR #68, opened 2019-02-19
Qsw 794 - Initial changes for Dashboard DB

claudio525:
Has to go with https://github.com/ucgmsim/qcore/pull/74


dashboard/DashboardDB.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+        ).fetchall()
+
+        conn.close()
+        return [SQueueEntry(*result) for result in results]
+
+    def _create_queue_table(self, cursor, hpc: const.HPC):
+        # Add latest table
+        cursor.execute(
+            """CREATE TABLE {}(
```
jasonmotha: CREATE TABLE IF NOT EXISTS 

So it doesn't break if it already exists even though technically it never should.

dashboard/DashboardDB.py
```diff
@@ -0,0 +1,344 @@
...
+        """Updates the daily core hours usage.
+
+        The core hour usage for a day is calculated as
+        last saved (for current day) TOTAL_CORE_USAGE - current total_core_usage.
+        """
+        # Do nothing if total_core_usage is None
+        if total_core_usage is None:
+            return
+
+        conn = sql.connect(self.db_file)
```
jasonmotha: Can we make a setup and teardown with and a 'with'? Seems like that would be the best practice.

sungeunbae: @jasonmotha  Can you show an example?

claudio525: have added this in commit   https://github.com/ucgmsim/slurm_gm_workflow/pull/68/commits/6c6224e377c429bcf175c4f0adeb242552af6a75

PR #71, opened 2019-02-20
Cross platform submission

jpa198:


estimation/model.py
```diff
@@ -225,7 +225,7 @@ def save_model(self, output_file: str):
 
         self._model.save(output_file)
 
-        with h5py.File(output_file, "r+") as f:
+        with h5py.File(output_file, "r") as f:
```
jasonmotha: Is this a merge issue?

claudio525: This change is incorrect, this is the save function so it requires write access. So this change should be reverted.

scripts/cybershake/auto_submit.py
```diff
@@ -21,6 +21,15 @@
 default_n_runs = 12
 default_1d_mod = "/nesi/transit/nesi00213/VelocityModel/Mod-1D/Cant1D_v2-midQ_leer.1d"
 
+job_run_machine = {
+    const.ProcessType.EMOD3D.value: "maui",
+    const.ProcessType.merge_ts.value: "mahuika",
+    const.ProcessType.winbin_aio.value: "maui",
+    const.ProcessType.HF.value: "maui",
+    const.ProcessType.BB.value: "maui",
+    const.ProcessType.IM_calculation.value: "maui",
```
claudio525: In qcore constants there is an enum for the different HPC systems, might be worth using that

scripts/submit_sim_imcalc.py
```diff
@@ -6,9 +6,10 @@
 
 import qcore.constants as const
 from qcore import utils, shared
+from qcore.config import host
 from typing import Dict
 from estimation.estimate_wct import est_IM_chours_single
-from shared_workflow.shared import submit_sl_script, set_wct, confirm, resolve_header
+from shared_workflow.shared import submit_sl_script, set_wct, confirm, get_partition, resolve_header
```
claudio525: get_partition doesn't appear to be used?

PR #81, opened 2019-02-26
Params to root

ttgenius:
1.deleted dt, hf_dt etc in cybershake_config
2.make all use of dt, hf_dt etc from root_defaults
3.testing results on maui: fine with manual install; 
    MPID_Init(246).......: channel initialization failed for merge_ts with auto_submit


jasonmotha:
Did you check the submit_bb & submit_hf scripts too? @ttgenius

ttgenius:
Yes they use root_defaults

PR #82, opened 2019-02-27
Robin changes

viktor76525:


scripts/bb_sim.py
```diff
@@ -266,14 +278,16 @@ def unfinished():
             "highpass",
         )
         lf_acc[:, c] = bwfilter(
-            ampdeamp(
```
jasonmotha: This one liner becomes hard to follow once split into mulitple lines.

shot0829: i think the typical 
```if statment:
else:
``` 
would probably be more readable

PR #83, opened 2019-02-27
Lf checkpointing

jpa198:


scripts/submit_emod3d.py
```diff
@@ -25,13 +25,19 @@ def write_sl_script(
...
+    nt=None,
 ):
     """Populates the template and writes the resulting slurm script to file"""
     workflow_config = load_config.load(
         os.path.dirname(os.path.realpath(__file__)), "workflow_config.json"
     )
 
-    set_runparams.create_run_params(srf_name, workflow_config=workflow_config)
+    if nt:
+        steps_per_checkpoint = int(nt/(60.0*convert_time_to_hours(run_time)) * 10)
```
jasonmotha: Spacing and ordering.

10 minutes should be a constant. i.e. CHECKPOINT_DURATION

scripts/submit_emod3d.py
```diff
@@ -116,6 +122,7 @@ def main(args):
             run_time=wct,
             nb_cpus=n_cores,
             machine=args.machine,
+            nt=int(float(params.sim_duration) / float(params.dt)),
```
jasonmotha: Make a shared function to get NT from params. Or a parameter.

PR #86, opened 2019-02-28
Jinjia

ttgenius:
Converted submission scripts to jinja except winbin_aio in submit_post_emod3d
tested with manual and auto install/submission on maui


scripts/submit_bb.py
```diff
@@ -12,6 +14,28 @@
 default_wct = "00:30:00"
 
 
+def generate_context(
```
claudio525: Instead of creating a new generate_context (and generate_header) function for every submit script, create a single shared function. As these functions all do the same thing, the only difference are the arguments, use **kwargs to deal with that. 

If you are not sure what I mean, let me know, I'm happy to explain.

jasonmotha: Can we still use kwargs when setting the function_parameter name? I wasn't sure about that otherwise I would have suggested it.

scripts/submit_empirical.py
```diff
@@ -1,37 +1,47 @@
...
+    return run_name.split("_")[0]
 
 
 def generate_header(template_path, account, np):
     j2_env = Environment(loader=FileSystemLoader(template_path), trim_blocks=True)
-    header = j2_env.get_template('slurm_header.cfg').render(version='slurm', job_description='Empirical Engine',
-                                                            job_name='empirical', account=account, nb_cpus=np,
-                                                            wallclock_limit="00:30:00", exe_time='%j', mail='',
-                                                            memory='2G', additional_lines='')
+    header = j2_env.get_template("slurm_header.cfg").render(
```
jasonmotha: Use the generate header function in shared.

scripts/submit_emod3d.py
```diff
@@ -109,11 +117,8 @@ def main(args):
         binary_path = binary_version.get_lf_bin(
             params.emod3d.emod3d_version, target_qconfig["tools_dir"]
         )
-
         steps_per_checkpoint = int(
-            get_nt(params)
-            / (60.0 * convert_time_to_hours(est_run_time))
```
claudio525: Any reason for removing the convert_time_to_hours call?

ttgenius:
est_run_time already in hours. Using covert_time_to_hours results in run
time error

scripts/submit_emod3d.py
ttgenius: est_run_time already in hours. Using covert_time_to_hours results in run time error

claudio525: :+1:

PR #88, opened 2019-02-28
Set Maui and Mahuika to load python 3 environment on login

jpa198:


install/load_default_modules.sh
```diff
@@ -2,25 +2,19 @@
...
     export PATH=/nesi/project/nesi00213/opt/mahuika/gmt/5.4.4/bin:$PATH
     export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/nesi/project/nesi00213/opt/mahuika/gmt/5.4.4/lib64
 elif [[ `hostname` =~ 'maui' ]] || [[ $HOSTNAME =~ 'ni' ]];then 
-    module load cray-python/2.7.15.1
+    load_python3_maui
     #python libs
-    export PYTHONPATH=/nesi/project/nesi00213/opt/maui/python-packages/lib/python2.7/site-packages:$PYTHONPATH
     export PATH=/nesi/project/nesi00213/opt/maui/python-packages/bin:$PATH
-    #qcore
-    export PYTHONPATH=/nesi/project/nesi00213/opt/maui/qcore:$PYTHONPATH
```
shot0829: I'm pretty sure the qcore libs are not 'installed' in the virtual environment, yet.
so to remove this line, we either have to setup the libs properly;
or we kept this line, but modify it so both maui/mahuika uses the same path (since the load_config in qcore now determines the machine, and we no longer need two different qcore libs)

jasonmotha: The load_python3 line sets up the python path correctly for the virtual environment.

shot0829: But qcore is not installed in the virtual_env.

jasonmotha:
qcore removed from opt folder. And pulled latest master in deploy

all packages installed to /opt/python-packages installed to the relevant virtual-env

Requirements.txt generated for each virtual_env from pip

jpa198:
Master pulled into deploy
qcore installed into each virtual_env
requirements.txt generated for each virtual_env and installed
qcore moved from opt (to opt/qcore_backup)

PR #89, opened 2019-03-01
Qsw 862 - Dashboard

claudio525:


dashboard/DashboardDB.py
```diff
@@ -116,7 +155,48 @@ def update_squeue(self, squeue_entries: Iterable[SQueueEntry], hpc: const.HPC):
...
+                    (
+                        entry.id,
+                        entry.name,
+                        entry.int_value_1,
+                        entry.int_value_2,
+                        datetime.now(),
+                    ),
+                )
 
     def get_sqeue_entries(self, hpc: const.HPC):
```
jasonmotha: spelling

dashboard/run_data_collection.py
```diff
@@ -0,0 +1,232 @@
...
+    # Update dashboard db
+    if rt_ch_output:
+        dashboard_db.update_daily_chours_usage(
+            parse_total_chours_usage(rt_ch_output, hpc), hpc
+        )
+    else:
+        return False
+
+    # Squeue
+    sq_output = run_cmd(user, hpc.value, "squeue")
```
jasonmotha: -A nesi00213

claudio525: I recon log everything for now (have to include the account in the results though) and then add a filter option to the table, which I think dash makes quite easy

dashboard/DashboardDB.py
claudio525: fixed

PR #90, opened 2019-03-01
Syntax hot fix

shot0829:
as well as other small fixs


shot0829:
take note that the ability to skip merge_ts makes it marked as 'completed' in the db.
a proper fix/implementation should be carried out next sprint.

PR #93, opened 2019-03-04
Test emod3d extend

jpa198:
Changed to check Rlog for line about writing seis and xyts files, and only check for the files if the line is there

### DIFF
--- a/scripts/test_emod3d.sh
+++ b/scripts/test_emod3d.sh
@@ -51,16 +51,20 @@ do
     seisFile=`echo $rootFile | sed 's/\(.*\)-/\1_seis-/'`
     xytsFile=`echo $rootFile | sed 's/\(.*\)-/\1_xyts-/'`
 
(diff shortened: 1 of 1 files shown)


scripts/test_emod3d.sh
```diff
@@ -79,9 +83,9 @@ then
 fi
 
 # EMOD3D is not considered to be completed if:
-#   Any Rlog does not have IS FINISHED
-#   There is a seis file without an equally indexed xyts file or vice versa
-#   Any seis file is unable to be loaded by LFSeis
+#  - Any Rlog does not have IS FINISHED
```
jasonmotha: does not contain "IS FINISHED"

scripts/test_emod3d.sh
```diff
@@ -79,9 +83,9 @@ then
 fi
 
 # EMOD3D is not considered to be completed if:
```
jasonmotha: I think this is easier to read / understand if we invert the logic. 'EMOD3D is considered to be complete if'

PR #95, opened 2019-03-05
Obs im calc

jpa198:


scripts/obs_im_calc.sl
```diff
@@ -0,0 +1,32 @@
...
+export PYTHONPATH=$gmsim/qcore:/$PYTHONPATH:$IMPATH
+
+script_start=`date`
+echo "script started running at: $script_start"
+
+obs_dirs=$1
+
+echo ___calculating observed____
+
+for D in `find $obs_dirs -type d`
```
jasonmotha: doesn't for D in $(ls) work faster?

scripts/obs_im_calc.sl
```diff
@@ -0,0 +1,32 @@
...
+do
+    if [[  `find $D -maxdepth 3 -mindepth 3 -name "accBB" | wc -l` -ge 1 ]]
```
jasonmotha: Potentiallly could use ls with '*/*/accBB' too?

scripts/calc_rrups_single.sl
```diff
@@ -0,0 +1,60 @@
...
+else
+    echo "rrup file already present: ${OUT_DIR}/rrup_${REL_NAME}.csv"
+    echo "Checking that there are enough rrups in it"
+fi
+
+if [[ -f ${OUT_DIR}/rrup_${REL_NAME}.csv ]] && [[ `wc -l ${OUT_DIR}/rrup_${REL_NAME}.csv` == `wc -l ${FD}` ]]
+then
+    echo "python $gmsim/workflow/scripts/management/update_mgmt_db.py $MGMT_DB_LOC rrup completed --run_name $REL_NAME --force" >> ${MGMT_DB_LOC}/mgmt_db_queue/${timestamp}\_${SLURM_JOBID}
+else
+    if [[ -f ${OUT_DIR}/rrup_${REL_NAME}.csv ]]
```
jasonmotha: Use nested if if, rather than checking for the same conditional.

PR #100, opened 2019-03-12
QSW 879 - Environments

claudio525:
Also renamed install to install_workflow.


install_workflow/README.md
```diff
@@ -0,0 +1,58 @@
+## Environment installation
+
+Requirements: Setup github SSH keys for maui, 
+[guide](https://help.github.com/en/articles/connecting-to-github-with-ssh) for how to do this. 
+
+A user specific environment of workflow, qcore, and IMCalc can be created as follows:
```
jasonmotha: Pre-processing and empirical engine too please.

scripts/cybershake/auto_submit.py
```diff
@@ -80,6 +80,7 @@ def submit_task(
             {"submit_time": submitted_time},
         )
 
+    # Merge ts
```
jasonmotha: Does it really need the comment if the first line checks if it is mergets?

claudio525: Haha probably not, but most of the others if have it.

PR #105, opened 2019-03-18
Im calc checkpoint extn

jpa198:


scripts/im_calc_checkpoint.py
```diff
@@ -16,77 +24,92 @@
...
-    for srf in srf_files[:]:
-        srf_name = srf.split("/")[-1].split(".")[0]
-        output_path = os.path.join(output_dir, srf_name + CSV_SUFFIX)
-        if os.path.isfile(output_path):
-            srf_files.remove(srf)
-    return srf_files
+        station_dir = os.path.join(output_dir, "stations")
+        if os.path.isdir(station_dir) and not event_name:
+            station_files = glob.glob1(station_dir, CSV_PATTERN)
+            if 0 < station_count != len(station_files):
```
jasonmotha: This either needs logic expansion or a comment. It's not intuitive what it does. I prefer the former.

PR #108, opened 2019-03-19
Fixed issues with header-body concatenation and line count return value

jpa198:


templates/sim_im_calc.sl.template
```diff
@@ -20,7 +22,7 @@ if [[ ! -d {{sim_dir}}/IM_calc ]]; then
 fi
 
 FD=$(getFromYaml "{{sim_dir}}/sim_params.yaml" FD_STATLIST)
-python $gmsim/workflow/scripts/im_calc_checkpoint.py {{sim_dir}}/IM_calc/ `wc -l ${FD}` --simulated
+python $gmsim/workflow/scripts/im_calc_checkpoint.py {{sim_dir}}/IM_calc/ `wc -l < ${FD}` --simulated
```
claudio525: Whats happening here? Am I missing something?

jpa198: If you pass the file name as an argument it will print it after the length of the file. If you pipe in the files contents it only shows the length of the file.

jpa198: image

claudio525: Ahhhhhhhhhhhhh, quality bash magic haha

PR #110, opened 2019-03-20
bb_sim and hf_sim now only parse known args. Any extras are ignored.

jpa198:
All settings in params.bb and params.hf are now added as parameters in
their respective slurm scripts
On install all settings under the hf and bb key of the sim_params yaml
are added to the sim_params.yaml file in the realisation directory


jpa198:
Is linked with pull request of same name in Pre Processing.
Has been tested on Maui and seems to work.

PR #116, opened 2019-03-22
Add slurm script location

jpa198:


scripts/submit_emod3d.py
```diff
@@ -71,20 +72,21 @@ def write_sl_script(
         job_description="emod3d slurm script",
         additional_lines="#SBATCH --hint=nomultithread",
         target_host=machine,
+        write_directory=write_directory,
     )
 
     fname_slurm_script = "run_emod3d_%s_%s.sl" % (srf_name, const.timestamp)
+    fname_slurm_script = os.path.abspath(
```
jasonmotha: Less confusing to merge these two lines

jasonmotha: Why abspath?

shared_workflow/shared.py
```diff
@@ -128,21 +128,9 @@ def generate_context(simulation_dir, template_path, **kwargs):
...
-    memory,
-    exe_time,
-    job_description,
-    partition=None,
-    additional_lines="",
-    template_path="slurm_header.cfg",
-    target_host=host,
-    mail="test@test.com",
-):
+def resolve_header(account, n_tasks, wallclock_limit, job_name, version, memory, exe_time, job_description,
```
jasonmotha: Black (as discussed)

scripts/submit_emod3d.py
jpa198: Returns the full path from root, so things like . and ~ are expanded

PR #117, opened 2019-03-24
QSW - 916 Added SVR model for extrapolation of core hours estimation

claudio525:
Goes with https://github.com/ucgmsim/qcore/pull/95


scripts/submit_emod3d.py
```diff
@@ -38,10 +39,6 @@ def write_sl_script(
     steps_per_checkpoint=None,
 ):
     """Populates the template and writes the resulting slurm script to file"""
-    workflow_config = load_config.load(
-        os.path.dirname(os.path.realpath(__file__)), "workflow_config.json"
```
jpa198: You're removing this from submit_emod3d, but adding it to submit_hf?

shared_workflow/load_config.py
```diff
@@ -3,7 +3,7 @@
 
 # Shouldn't this be in shared or qcore?
 def load(
-    directory=os.path.dirname(os.path.abspath(__file__)),
+    directory=os.path.join(os.path.dirname(os.path.abspath(__file__)), "../scripts"),
```
jpa198: Use `os.path.abspath(os.path.join(os.path.dirname(__file__), "../scripts"))` to get the same effect, but resolve the `..`?

PR #130, opened 2019-03-26
Old data

ttgenius:
added class for collecting old core hour usage data


.travis.yml
```diff
@@ -43,6 +43,6 @@ script:
     python setup.py install;
     cd /home/root/git/slurm_gm_workflow;
     export PYTHONPATH=$PYTHONPATH:/home/root/git/slurm_gm_workflow;
-    pytest -vs --ignore testing/test_manual_install --ignore testing/test_dashboard;"
+    pytest -vs --ignore testing/test_manual_install;"
```
jpa198: 👍

dashboard/run_old_data_collection.py
```diff
@@ -0,0 +1,130 @@
...
+    parser.add_argument("dashboard_db", help="The dashboard db to use")
+    parser.add_argument(
+        "--update_interval",
+        help="Interval between data collection (seconds)",
+        default=30,
+    )
+    parser.add_argument(
+        "--hpc",
+        choices=HPCS,
+        help="Specify the HPCs on which to collect data, defaults to all HPCs",
```
jpa198: Should have nargs="+" set to allow specifying multiple, or reword to be singular.

dashboard/run_old_data_collection.py
```diff
@@ -0,0 +1,130 @@
...
+        # Collect the data
+        print("{} - Collecting data from HPC {}".format(datetime.now(), hpcs))
+        for hpc in hpcs:
+            collect_old_data(dashboard_db, hpc, login_user, users, day_shift)
+            print("{} - Done".format(datetime.now()))
+
+
+def main(args):
+    hpc = (
+        const.HPC(args.hpc)
```
jpa198: Should be [const.HPC(args.hpc)], so that you're always passing a list into run_old_collection, and don't need line 71?

ttgenius: I tried this before, but then the 'hpc' would not work for function "collect_old_data"

dashboard/run_old_data_collection.py
ttgenius: --hpc is default to mutilple, if you decide to specify -hpc, only one hpc should be specified

PR #132, opened 2019-03-26
QSW 962 - End-to-End test Checkpointing

claudio525:


e2e_tests/E2ETests.py
```diff
@@ -299,43 +303,98 @@ def run_auto_submit(self, timeout: int = 10, sleep_time: int = 10):
...
+        """Looks for any running task of the specified process types
+        and attempts to cancel one of each.
+        """
+        # Get all running jobs in the mgmt db
+        proc_types_int = [proc_type.value for proc_type in proc_types]
+        with connect_db_ctx(sim_struct.get_mgmt_db(self.stage_dir)) as cur:
+            entries = cur.execute(
+                "SELECT proc_type, job_id  FROM state "
+                "WHERE status == 3 AND proc_type IN ("
+                + "?," * (len(proc_types_int) - 1)
```
jasonmotha: Arguably nicer to use python slicing.

jpa198: Could use "IN ({}?)".format("?,"*(len(proc_types_int) - 1)) ?

jasonmotha: That's nicer still. Don't have to do string appending this way.

PR #137, opened 2019-03-28
Tar task

jpa198:


scripts/clean_up.sl
```diff
@@ -0,0 +1,39 @@
...
+start_time=`date +${runtime_fmt}`
+echo ___cleaning up___
+echo "python $gmsim/workflow/scripts/management/update_mgmt_db.py $MGMT_DB_LOC clean_up running --run_name $SRF_NAME --job $SLURM_JOBID" >> $MGMT_DB_LOC/mgmt_db_queue/$timestamp\_$SLURM_JOBID
+
+res = `time python $gmsim/workflow/scripts/clean_up.py $SIM_DIR`
+
+end_time=`date +$runtime_fmt`
+echo $end_time
+
+if [[ $? == 0 ]]; then
```
jasonmotha: as discussed

scripts/cybershake/auto_submit.py
```diff
@@ -104,7 +106,9 @@ def submit_task(
         # skipping winbin_aio if running binary mode
         if binary_mode:
             # update the mgmt_db
-            update_mgmt_db.update_db(db, "winbin_aio", "completed", run_name=run_name)
+            update_mgmt_db.update_db(
+                db, "winbin_aio", const.State.completed.value[1], run_name=run_name
```
jasonmotha: Can update the process name too

scripts/cybershake/auto_submit.py
```diff
@@ -301,11 +331,32 @@ def main():
...
+                )
+                task_num = task_num + 1
+                continue
+            elif slurm_query_status.is_task_complete(
+                [
+                    const.ProcessType.clean_up.value,
+                    run_name,
+                    const.State.completed.value[1],
+                ],
+                slurm_query_status.get_db_tasks_to_be_run(db),
```
jasonmotha: Rather than querying here, leave the second argument as a optional parameter and run the command as the default if it doesn't exist. 

I think it's better separation of logic.

scripts/management/slurm_query_status.py
```diff
@@ -23,14 +23,19 @@
...
 def get_queued_tasks(user=None):
-    if user != None:
-        cmd = "squeue -A nesi00213 -o '%A %t' -h" + " -u " + user
-    else:
-        cmd = "squeue -A nesi00213 -o '%A %t' -h"
-    process = Popen(shlex.split(cmd), stdout=PIPE, encoding="utf-8")
-    (output, err) = process.communicate()
-    exit_code = process.wait()
-    return output
+    output_list = []
```
jasonmotha: This needs a to-do to change to treating mahuika and maui jobs separately and linked in to QSW-912

scripts/management/slurm_query_status.py
```diff
@@ -145,17 +150,34 @@ def check_dependancy_met(task, task_list):
         return is_task_complete(LF_task, task_list) and is_task_complete(
             HF_task, task_list
         )
+
+    if process is Process.clean_up:
```
jasonmotha: Doesn't function 100% as desired but plot_flag should help with that.

templates/post_emod3d_merge_ts.sl.template
```diff
@@ -21,6 +21,7 @@ if [[ ! "$(ls -A {{lf_sim_dir}}/OutBin)" ]]; then
      tar -xvf {{sim_dir}}/submission.tar
```
jasonmotha: does it still need submission.tar?

jpa198: Currently the tar files are gzip compressed (extension will reflect that in the next push).
Options are to extract the entire archive, then tar the old and new files together into a new archive, or to remove compression and append them to the old archive.
Will go with option 1 unless someone says otherwise.

scripts/management/slurm_query_status.py
jpa198: How should it function?

jasonmotha: If merge ts isn't complete it won't 'clean-up'. 
If we are skipping merge-ts (via no-merge-ts or plotting) then it should clean up by default unless no-clean up is set.

templates/post_emod3d_merge_ts.sl.template
jasonmotha: Yeah, they don't need to be compressed. Binaries especially are already compressed and then the slurm files are so small the compression doesn't matter.

PR #140, opened 2019-03-31
Dash alert

ttgenius:
alert when update time gap exceeds 5 minutes
1. error_table will be created when creating the database
2. when update time exceeds time gap limit, error will be inserted to db
3. web app check if there's record in the error table. If there's is, alert.
4. if error resolved, err record will be deleted from the error table


dashboard/run_dashboard_app.py
```diff
@@ -126,6 +135,19 @@ def update_maui_total_chours(n):
...
+              [Input("interval_comp", "n_intervals")])
+def display_confirm(n):
+    return app.db.get_collection_err(const.HPC.maui) is not None
+
+
+@app.callback(Output('output-confirm', 'children'),
+              [Input('confirm', 'submit_n_clicks')])
+def update_output(submit_n_clicks):
+    if submit_n_clicks:
+        return None
```
jpa198: Does this do anything? A function without a return statement returns a None value anyway.

ttgenius: it returns None when you click the options in the alert box

dashboard/run_dashboard_app.py
```diff
@@ -34,23 +35,20 @@
...
-                    html.H5("Daily core hour usage"),
-                    dcc.Graph(id="maui_daily_chours"),
-                    html.H5("Total core hour usage"),
-                    dcc.Graph(id="maui_total_chours"),
-                    dcc.Interval(id="interval_comp", interval=10 * 1000, n_intervals=0),
-                    html.H5("maui_daily_inodes"),
-                    dcc.Graph(id="maui_daily_inodes"),
-                ]
-            ),
+            html.Div(id="err"),
```
jpa198: This has its content set when display_err is called?

ttgenius: yes, display_error(n) function output html_text to the html_div with id="err"

dashboard/run_dashboard_app.py
```diff
@@ -126,6 +124,14 @@ def update_maui_total_chours(n):
     return fig
 
 
+@app.callback(Output('err', 'children'),
+              [Input("interval_comp", "n_intervals")])
+def display_err(n):
+    """Displays data collection error when the gap between update_times exceeds acceptable limit"""
+    if app.db.get_collection_err(const.HPC.maui) is not None:
+        return html.Plaintext("Data collection err")
```
claudio525: maybe spell the word out completely?

dashboard/DashboardDB.py
```diff
@@ -334,6 +367,27 @@ def get_user_chours(self, hpc: const.HPC, username: str):
 
         return [UserChEntry(*result) for result in results]
 
+    def check_update_time(self, last_update_time_string, current_update_time):
+        """
+        Checks whether the time gap between update times exceeds the idling time limit
+        if exceeds, regards as a collection error.
+        :param last_update_time_string:
```
claudio525: Specifying the parameters without a proper description isn't overly useful, so maybe remove these and add use type hinting if you want to specify the types?

dashboard/DashboardDB.py
```diff
@@ -106,7 +111,7 @@ def update_daily_chours_usage(
 
         with self.get_cursor(self.db_file) as cursor:
             row = cursor.execute(
-                "SELECT CORE_HOURS_USED, TOTAL_CORE_HOURS FROM {} WHERE DAY == ?;".format(
+                "SELECT CORE_HOURS_USED, TOTAL_CORE_HOURS, UPDATE_TIME FROM {} WHERE DAY == ?;".format(
```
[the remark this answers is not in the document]
ttgenius: originally , updatetime check is performed here as core_hour_usage function is first get called when collection data. But nesi had issue when using nn_corehour_usage last Friday, so I shifted the check to update_user_chours funciton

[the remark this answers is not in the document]
claudio525: Makes sense. Since its no longer need here, maybe just remove it.

dashboard/DashboardDB.py
ttgenius: updated

PR #142, opened 2019-04-01
app now plots physical core hours by default

ttgenius:
app now plots physical core hours by default


ttgenius:
all changes to phycial chours are in branch phy_chours

PR #147, opened 2019-04-04
Sreport chours

ttgenius:
use sreport for getting daily and total core hours usage


jasonmotha:
I'm happy but haven't approved as I'm not as familiar with the details of it.

PR #149, opened 2019-04-08
Mahuika chours

ttgenius:
display mahuika along with maui core hours from 188 days ago (text and plot)


dashboard/run_data_collection.py
```diff
@@ -47,6 +47,7 @@ class DataCollector:
 
     utc_time_format = "%m/%d/%y-%H:%M:%S"
     utc_time_gap = datetime.now() - datetime.utcnow()
+    total_start_time = "09/25/18-12:00:00"
```
claudio525: why this date? Doesn't our allocation start from beginning of june?

dashboard/run_data_collection.py
```diff
@@ -258,7 +257,8 @@ def _parse_squeue(self, lines: Iterable[str]):
 
         return entries
 
-    def _parse_chours_usage(self, ch_lines: List):
+    @staticmethod
+    def parse_chours_usage(ch_lines: List):
```
claudio525: Why make this public?

dashboard/run_data_collection.py
ttgenius: crop out the zeros

dashboard/run_data_collection.py
ttgenius: it's used by run_old_data _colleciton

PR #151, opened 2019-04-09
QSW 1057 - Automated workflow improvements

claudio525:
Note: The MgmtDB class is an intial attempt at moving the mgmt db functionality into a single location vs having it spread over lots of scripts. However, this is not complete, more refactoring needs to be done, a jira task has been created for this.


scripts/cybershake/auto_submit.py
```diff
@@ -19,78 +25,168 @@
...
+                            const.Status(db_task.status).str_value,
+                            queue_status,
+                        )
+                    )
+                    shared.add_to_queue(
+                        queue_folder, db_task.run_name, db_task.proc_type, queue_status
+                    )
+        # Only reset if there is no entry on the mgmt queue for this
+        # realisation/proc combination
+        # Ignore cleanup for now as it runs on mahuika
```
jpa198: Merge_ts also runs on Mahuika and should be reenabled soon?

scripts/cybershake/add_to_mgmt_queue.py
```diff
@@ -0,0 +1,25 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("queue_folder", type=str, help="Mgmt db queue folder")
+    parser.add_argument("run_name", type=str, help="The realisation/run name")
+    parser.add_argument(
+        "proc_type", type=int, help="The integer value for the process type"
+    )
+    parser.add_argument("status", type=int, help="The integer value for the status")
```
jasonmotha: Any reason you used integer values rather than strings? 
I feel like integer values expose too much of the magic behind the curtain.

scripts/cybershake/auto_submit.py
```diff
@@ -19,78 +25,168 @@
...
+    const.ProcessType.BB: const.HPC.maui,
+    const.ProcessType.IM_calculation: const.HPC.maui,
+    const.ProcessType.clean_up: const.HPC.mahuika,
 }
 
+SLURM_TO_STATUS_DICT = {"R": 3, "PD": 2, "CG": 3}
+
+
+def get_queued_tasks(user=None, machine=const.HPC.maui):
+    # TODO: Treat Maui and Mahuika jobs seperately. See QSW-912
```
jasonmotha: Remove TODO

scripts/cybershake/queue_monitor.py
```diff
@@ -0,0 +1,91 @@
...
+            )
+        )
+
+    return queue_entries
+
+
+def main(args):
+    mgmt_db = MgmtDB(sim_struct.get_mgmt_db(args.root_folder))
+    queue_folder = sim_struct.get_mgmt_db_queue(args.root_folder)
+
```
jasonmotha: Probably should print instructions on how to close file. Or have an end condition for the while.

jpa198: Add a signal handler to ensure graceful close on ctrl+c?

scripts/cybershake/add_to_mgmt_queue.py
claudio525: This isn't really a user facing script, its only used from within the slurm scripts.

jpa198: The magic numbers for this in the slurm scripts isn't the best either

README.md
```diff
@@ -1,7 +1,85 @@
...
+```
+
+and list.txt is a list of the faults to run, along with the number of realisations, e.g.
+```
+Hossack 4r
+RepongaereF4 10r
+```
+
+### Running
+To run the installed simulations, navigate to the rootDir and run
```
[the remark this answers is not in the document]
claudio525: Yeah should be able to run it from anywhere. I just always do it that way.

scripts/cybershake/add_to_mgmt_queue.py
jasonmotha: I think that if there are changes to the IDs. Say to avoid giving pre-processing tasks negative numbers. Then it's not obvious (from a search) that this code has to also change. If we are explicit with str names, it should be much more obvious when something 'wrong' is happening.

claudio525: Hmh I don't think we should ever change the IDs, since they are IDs and don't need to be in order. I prefer magic number to magic string, but I'll change it to strings.

shot0829: i would personally prefer strings as well, this will make debug or any manual override a lot easier.

scripts/cybershake/auto_submit.py
```diff
@@ -19,78 +25,165 @@
...
+
+    process = Popen(shlex.split(cmd), stdout=PIPE, encoding="utf-8")
+    (output, err) = process.communicate()
+    process.wait()
+
+    output_list = list(filter(None, output.split("\n")[1:]))
+    return output_list
+
+
+def check_queue(queue_folder: str, run_name: str, proc_type: int):
```
shot0829: would probably be less confusing if we change the name a little bit, since now we have two 'queue': slurm queue and a PIPE queue for updaing management DB

scripts/cybershake/auto_submit.py
```diff
@@ -135,179 +229,109 @@ def submit_task(
...
-    mgmt_db_location = args.run_folder
-    db = connect_db(mgmt_db_location)
+def main(args):
+    root_folder = args.root_folder
+    queue_folder = sim_struct.get_mgmt_db_queue(root_folder)
+    mgmt_db = MgmtDB(sim_struct.get_mgmt_db(root_folder))
 
     # Default values
-    oneD_mod, hf_vs30_ref, binary_mode, hf_seed = default_1d_mod, None, True, None
+    oneD_mod, hf_vs30_ref, binary_mode, hf_seed = DEFAULT_1D_MOD, None, True, None
```
shot0829: oneD_mod, hf_vs30ref can be removed, since they are not used in anywhere and cannot be override. 
(simulation reads the value from params.yaml)

shared_workflow/shared.py
```diff
@@ -612,34 +614,53 @@ def submit_sl_script(
...
-        cmd_name = os.path.join(db_queue_path, "%s_%s_q" % (timestamp, jobid))
-        cmd = (
-            "python $gmsim/workflow/scripts/management/"
-            "update_mgmt_db.py {} {} {} --run_name {}  --job {}".format(
-                mgmt_db_loc, process, status, srf_name, jobid
+def add_to_queue(
+    queue_folder: str,
+    run_name: str,
+    proc_type: int,
+    status: int,
```
shot0829: have you run test after most recent commit?
this function still takes INT for status, but there is no enum or any translation from STR in "add_to_mgmt_queue.py"

shot0829: have you run test after most recent commit?
this function still takes INT for status, but there is no enum or any translation from STR in "add_to_mgmt_queue.py"

scripts/submit_bb.py
```diff
@@ -19,9 +20,8 @@
 default_wct = "00:30:00"
 
 
-def main(args):
+def main(args, est_model=None):
```
shot0829: for consistency, change this to what submit_emod3d.py has?
stricter but more clear of declaration of usage of variables.

scripts/submit_emod3d.py
```diff
@@ -23,7 +24,7 @@
 CHECKPOINT_DURATION = 10
 
 
-def main(args):
+def main(args: argparse.Namespace, est_model: est.EstModel = None):
```
shot0829: for reference

shared_workflow/shared.py
```diff
@@ -612,34 +614,53 @@ def submit_sl_script(
...
-    """Adds the command to update the mgmt db to the queue"""
-    # Update the db
-    if mgmt_db_loc is not None and os.path.isdir(mgmt_db_loc):
-        db_queue_path = os.path.join(mgmt_db_loc, "mgmt_db_queue")
-        cmd_name = os.path.join(db_queue_path, "%s_%s_q" % (timestamp, jobid))
-        cmd = (
-            "python $gmsim/workflow/scripts/management/"
-            "update_mgmt_db.py {} {} {} --run_name {}  --job {}".format(
-                mgmt_db_loc, process, status, srf_name, jobid
+def add_to_queue(
```
shot0829: any reason to put this function into the shared.py? 
because by doing so we create extra dependencies on a shared lib.

shared_workflow/shared.py
```diff
@@ -7,20 +7,24 @@
...
+from datetime import datetime
 
 from jinja2 import Environment, FileSystemLoader
 
 from qcore import binary_version
 from qcore.config import host
 import qcore.constants as const
 from qcore.utils import load_sim_params
+from scripts.cybershake.queue_monitor import DATE_FORMAT as QUEUE_DATE_FORMAT
+from scripts.management.MgmtDB import MgmtDB
```
shot0829: as mentioned.

scripts/submit_bb.py
claudio525: Done

scripts/cybershake/add_to_mgmt_queue.py
```diff
@@ -13,7 +13,7 @@
     parser.add_argument(
         "proc_type", type=str, help="The string value of the process type."
     )
-    parser.add_argument("status", type=int, help="The integer value for the status")
+    parser.add_argument("status", type=str, help="The string value of the status")
```
jasonmotha: Last thing. I'd add the enum as pre-poulated options for this so it errors in case of a misspelling.

PR #153, opened 2019-04-12
Dont copy templates

jpa198:


shared_workflow/shared.py
```diff
@@ -25,6 +25,7 @@
 from qcore.utils import load_sim_params
 from scripts.cybershake.queue_monitor import DATE_FORMAT as QUEUE_DATE_FORMAT
 from scripts.management.MgmtDB import MgmtDB
+from shared_workflow.shared_defaults import recipe_dir
```
claudio525: Anyone know why its called recipe_dir and not template_dir?

shot0829: legacy. we used to have tons of 'recipe' that combines templates to create stuffs for user.

shot0829: we dont have a folder call recipe anymore, i dont think its too impactful to leave it as is, since it does make some sense; you follow a exact recipe to make something(food), or in our case, simulation.

PR #155, opened 2019-04-15
Fix install path and merge ts path

jpa198:


jpa198:
Needs to be deployed in parallel with EMOD3D pull request

PR #164, opened 2019-04-26
extended timeout

ttgenius:
dashboard crash fix: ssh cmd timeout too short
red font for error text


dashboard/run_dashboard_app.py
```diff
@@ -146,7 +146,8 @@ def display_err(n):
     """Displays data collection error when the gap between update_times exceeds acceptable limit"""
     if not check_update_time(app.db.get_update_time(const.HPC.maui)[0], datetime.now()):
         return html.Plaintext(
-            "Data collection error, check the error_table in database"
+            "Data collection error, check the error_table in database",
+            style={"color": "red", "font-size": 20},
```
jasonmotha: I reckon background-color: 'red' would be more appropriate.

ttgenius: update

PR #166, opened 2019-04-29
Pre-processing now makes vm_params.yaml and slurm workflow just updates

jpa198:
this yaml with the GRID* and MODEL_* parameters.


jpa198:
params_vel.yaml will need to be renamed to vm_params.yaml for old source generation to work with this

PR #168, opened 2019-04-29
Remove extra params yaml

jpa198:


install_workflow/create_config_file.py
```diff
@@ -19,7 +19,7 @@
   "bin_process_path": bin_process_path,
   "qcore_lib_path": qcore_path,
   "install_bb_name": "install_bb.py",
-  "params_vel": "params_vel.yaml",
+  "vm_params": "vm_params.yaml",
```
shot0829: Since we are adding dependencies towards qcore.constant,  i think its a good time to remove this parameter from this config.

scripts/install.py
```diff
@@ -114,21 +114,21 @@ def q_select_vel_model(vel_mod_dir):
 
     v_mod_ver_options = []
     for root, dirnames, filenames in os.walk(vel_mod_dir):
-        # returns the folder that contains params_vel.py
-        for filename in fnmatch.filter(filenames, defaults.params_vel):
+        # returns the folder that contains vm_params.py
+        for filename in fnmatch.filter(filenames, defaults.vm_params):
```
shot0829: Same issue. so instead of reading from the workflow.config, get it from qcore.constant?

shared_workflow/shared_defaults.py
```diff
@@ -8,7 +8,7 @@
 global_root = workflow_config["global_root"]
 bin_process_dir = os.path.join(global_root, 'workflow/scripts')
 emod3d_version = workflow_config["emod3d_version"]
-params_vel = workflow_config['params_vel']
+vm_params = workflow_config['vm_params']
```
shot0829: if dependencies are correctly moved to qcore.constant, this shouldnt be needed anymore.

jpa198:
> install.py is still using "defaults.params_vel"? which is loaded from workflow.config

Not that I can see? Can you comment on the line?

PR #180, opened 2019-05-16
Only run needed tasks

jpa198:


e2e_tests/E2ETests.py
```diff
@@ -385,6 +385,7 @@ def _run_auto(self, user: str, sleep_time: int = 10):
                         "Operational error while accessing database. "
                         "Retrying in {} seconds\n{}".format(sleep_time, ex)
                     )
+                    time.sleep(5)
```
jasonmotha: Isn't this a parameter?

scripts/cybershake/auto_submit.py
```diff
@@ -35,6 +35,10 @@
     const.ProcessType.HF: const.HPC.maui,
     const.ProcessType.BB: const.HPC.maui,
     const.ProcessType.IM_calculation: const.HPC.maui,
+    const.ProcessType.IM_plot: const.HPC.mahuika,
+    const.ProcessType.rrup: const.HPC.mahuika,
```
jasonmotha: A cool extension would be to run things that are platform agnostic on whichever platform has a shorter queue. Or to saturate both queues.

scripts/cybershake/auto_submit.py
```diff
@@ -608,6 +611,15 @@ def main(args, main_logger: Logger = workflow_logger.get_basic_logger()):
...
         args.n_runs = DEFAULT_N_RUNS
 
+    logger.debug("Processes to be run were: {}. Getting all required dependencies now.".format(args.tasks_to_run))
+    args.tasks_to_run = [const.ProcessType.get_by_name(proc) for proc in args.tasks_to_run]
+    for task in args.tasks_to_run:
+        logger.debug("Process {} in processes to be run, adding dependencies now.".format(task.str_value))
+        for proc_num in task.dependencies:
+            proc = const.ProcessType(proc_num)
+            if proc not in args.tasks_to_run:
+                args.tasks_to_run.append(proc)
```
jasonmotha: I presume this cascades well? I can't explicity see that here but presume that adding to the iterating list gives that cascading.

scripts/management/MgmtDB.py
```diff
@@ -96,14 +96,19 @@ def get_runnable_tasks(self, retry_max):
             Process.Verification.value,
         ]
 
+        if allowed_tasks is None:
+            allowed_tasks = list(const.ProcessType)
+        allowed_tasks = list(map(lambda x: str(x.value), allowed_tasks))
```
jasonmotha: List comprehension.

scripts/management/MgmtDB.py
```diff
@@ -96,14 +96,19 @@ def get_runnable_tasks(self, retry_max):
...
         with connect_db_ctx(self._db_file) as cur:
             db_tasks = cur.execute(
                 """SELECT proc_type, run_name, status_enum.state 
                           FROM status_enum, state 
                           WHERE state.status = status_enum.id
+                           AND proc_type IN ({})
                            AND (((status_enum.state = 'created' OR status_enum.state = 'failed')  
                                  AND state.retries <= ?)
-                            OR status_enum.state = 'completed')""",
+                            OR status_enum.state = 'completed')""".format(", ".join(allowed_tasks)),
```
jasonmotha: Use prepared queries rather than string formatting.

scripts/cybershake/auto_submit.py
jpa198: Yep, adding to the end of the list means the for loop is dynamic and only ends when we stop adding things

scripts/cybershake/auto_submit.py
jpa198: Sounds like it should have its own task

jasonmotha: Yeah. I'm just droning on.

PR #183, opened 2019-05-20
Lf/hf only bb

jpa198:
Would like Jason to review before merge


scripts/hf2bb.sl
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+# script version: slurm
+#
+# must be run with sbatch hf2bb.sl [realisationDirectory] [managementDBLocation]
+
+#SBATCH --cluster=Mahuika
```
jasonmotha: Probably wanna revise these slurm headers. Probably single core on mahuika.

jasonmotha: Does calc single rrup have 40 cores too - good to check while we are here?

scripts/hf2bb.sl
```diff
@@ -0,0 +1,58 @@
...
+
+end_time=`date +$runtime_fmt`
+echo $end_time
+
+timestamp=`date +%Y%m%d_%H%M%S`
+#test before update
+res=`$gmsim/workflow/scripts/test_bb.sh $REL_LOC `
+if [[ $? == 0 ]]; then
+    #passed
+    python $gmsim/workflow/scripts/cybershake/add_to_mgmt_queue.py $MGMT_DB_LOC/mgmt_db_queue $REL_NAME BB completed
```
jasonmotha: I think there is a separate task for creating new tasks for these in the mgmt db along with alternative dependancies for BB.

jpa198: So between this task and that one, these scripts should have entries in the qcore process type enum, with associated MgmtDB entries and autosubmit logic?

scripts/hf2bb.sl
jpa198: I've dropped the calc_rrups _single core count to 1.
Run times for AlpineF2K on Maui lander were lf2bb:32 mins, hf2bb: 3mins, so these hard coded times should be ok for everything

PR #188, opened 2019-05-21
Im calc deps

jpa198:


scripts/cybershake/auto_submit.py
```diff
@@ -184,6 +186,17 @@ def submit_task(
     submitted_time = datetime.now().strftime(const.METADATA_TIMESTAMP_FMT)
     log_file = os.path.join(sim_dir, "ch_log", const.METADATA_LOG_FILENAME)
 
+    def submit_sl_script(script_name, **kwargs):
```
claudio525: What about doing a partial definition after the last if, instead of defining a new function?
https://docs.python.org/3/library/functools.html#functools.partial

scripts/management/slurm_mgmt.db.sql
```diff
@@ -43,6 +43,8 @@ INSERT OR IGNORE INTO `proc_type_enum` (id,proc_type) VALUES (6,'IM_calculation'
 --INSERT OR IGNORE INTO `proc_type_enum` (id,proc_type) VALUES (9,'Empirical');
 --INSERT OR IGNORE INTO `proc_type_enum` (id,proc_type) VALUES (10,'Verification');
 INSERT OR IGNORE INTO `proc_type_enum` (id,proc_type) VALUES (11,'clean_up');
+INSERT OR IGNORE INTO `proc_type_enum` (id,proc_type) VALUES (12,'LF2BB');
```
jasonmotha: Can remove the above comments as these tasks will no longer attempt to be run.

scripts/cybershake/auto_submit.py
```diff
@@ -660,7 +660,33 @@ def main(args, main_logger: Logger = workflow_logger.get_basic_logger()):
...
+                const.ProcessType.BB in args.tasks_to_run,
+                const.ProcessType.HF2BB in args.tasks_to_run,
+                const.ProcessType.LF2BB in args.tasks_to_run,
+            ]
+        )
+        > 1
+    ):
+        logger.log(
+            workflow_logger.NOPRINTCRITICAL,
+            "Mutually exclusive tasks have been added. "
```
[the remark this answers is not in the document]
jpa198: parser.error raises an exception and causes the script to exit

jpa198:
End 2 end test ran normally

PR #189, opened 2019-05-24
WCT_with_fixes

shot0829:
a lot of small fixes and wct scale together


scripts/hf_sim.py
```diff
@@ -155,11 +158,14 @@ def random_seed():
...
     )
     try:
-        args = parser.parse_known_args()[0]
+        args = parser.parse_args()
     except SystemExit:
         # invalid arguments or -h
         comm.Abort()
 
+if args.version is not None:
+    args.sim_bin = binary_version.get_hf_binmod(args.version)
```
jasonmotha: Confusing behaviour

scripts/bb_sim.py
```diff
@@ -50,7 +50,7 @@
     )
 
     try:
-        args = parser.parse_known_args()[0]
+        args = parser.parse_args()
```
jasonmotha: Why?

scripts/submit_emod3d.py
```diff
@@ -60,6 +60,19 @@ def main(args: argparse.Namespace, est_model: est.EstModel = None, logger: Logge
             model,
             True,
         )
+        # scale up the est_run_time if it is a re-run (with check-pointing)
```
jasonmotha: Why only scale up if checkpointing files exist?

scripts/bb_sim.py
shot0829: reverting back to raising error when unexpected parameters provided.
so when the user accidentally provided typo or anything incorrect, error will be raised;
instead of just running it through it default values and never realize its incorrect.

scripts/hf_sim.py
shot0829: i can probably remove this line.
this changes was to band-aid fix when the submit_* code did not handle the version provided in gmsim template correctly.

shot0829: if the submit code, either auto_submit or submit_*. did not handle the 'version' parameter correctly, with the code here it will still run with the version provided.
instead of just using default version

scripts/submit_emod3d.py
shot0829: otherwise its either an 'on purpose' cancel or something is so wrong that it might not even be restart-able

scripts/hf_sim.py
claudio525: Yeah the binary to use is determined in submit_hf.py, so I think this should be removed. And in extension the version argument, since its not used anywhere else?

PR #192, opened 2019-05-30
Auto submit wrapper

jpa198:


scripts/management/MgmtDB.py
```diff
@@ -72,17 +73,19 @@ def close_conn(self):
...
 
-    def get_submitted_tasks(self):
+    def get_submitted_tasks(self, allowed_tasks=tuple(const.ProcessType)):
         """Gets all in progress tasks_to_run i.e. (running or queued)"""
+        allowed_tasks = [str(task.value) for task in allowed_tasks]
         with connect_db_ctx(self._db_file) as cur:
             result = cur.execute(
                 "SELECT run_name, proc_type, status, job_id, retries "
-                "FROM state WHERE status IN (2, 3)"
+                "FROM state WHERE status IN (2, 3) AND proc_type IN (?{})".format(",?"*(len(allowed_tasks)-1)),
```
jasonmotha: Do the join to make the table easier to read. Silly SQLite not having built in enums

jpa198:
Additional documentation available here: 
https://wiki.canterbury.ac.nz/display/QuakeCore/Auto+submit+wrapper

claudio525:
Given that this will be the new way of running the automated workflow, we should update the End to End test to use this, instead of auto_submit directly (happy for this to be a separate task). And did you run the e2e tests for this? 
Once E2E use the wrapper it might also be worth running a larger E2E test on it, given how large the change is?

jpa198:
E2E test using the wrapper instead of auto_submit/queue_monitor has run successfully

jpa198:
Also includes fix for the n_runs argument

jpa198:
End2end validation ran successfully in about 3:45

PR #196, opened 2019-06-04
added type error exception for none chours data & extended timeout

ttgenius:
2019-06-04 12:05:04.594923 - Collecting data from HPC [<HPC.maui: 'maui'>, <HPC.mahuika: 'mahuika'>]
The timeout of 60 expired for cmd sreport -n -t Hours cluster AccountUtilizationByUser Accounts=nesi00213 start=01/06/18-12:00:00 end=06/04/19-12:00:00.
Traceback (most recent call last):
  File "run_data_collection.py", line 389, in <module>
    main(args)
  File "run_data_collection.py", line 358, in main
    data_col.run(args.users)
  File "run_data_collection.py", line 97, in run
    self.collect_data(hpc, users)
  File "run_data_collection.py", line 145, in collect_data
    rt_total_ch = self.parse_chours_usage(rt_total_ch_output)
  File "run_data_collection.py", line 268, in parse_chours_usage
    return ch_lines[0].strip().split()[-2]
TypeError: 'NoneType' object is not subscriptable

### DIFF
--- a/dashboard/run_data_collection.py
+++ b/dashboard/run_data_collection.py
@@ -269,7 +269,7 @@ class DataCollector:
             return 0
         except ValueError:
             print("Failed to convert total core hours to integer.")
-       
+
     def _parse_quota(self, lines: Iterable[str]):
         """
         Gets quota usage from cmd and return as a list of QuotaEntry objects
@@ -366,7 +366,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--update_interval",
         help="Interval between data collection (seconds)",
-        default=30,
+        default=300,
     )
     parser.add_argument(
         "--hpc",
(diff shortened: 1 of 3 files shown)


ttgenius:
> Is it better to check the output rather than catch exceptions? Is that the only case where that exception occurs?
Agree. Better to check output. Yes this is the only place that output is not checked. Have updated

dashboard/run_data_collection.py
```diff
@@ -131,23 +131,21 @@ def collect_data(self, hpc: const.HPC, users: Iterable[str]):
...
                 PROJECT_ID, self.total_start_time, end_time
             ),
-            timeout=60,
         )
-        rt_total_ch = self.parse_chours_usage(rt_total_ch_output)
-
         if rt_total_ch_output:
-            self.dashboard_db.update_chours_usage(rt_daily_ch, rt_total_ch, hpc)
-        # Squeue, formatted to show full account name
+            rt_total_ch_output = self.parse_chours_usage(rt_total_ch_output)
```
jasonmotha: Parses twice now?

PR #197, opened 2019-06-04
New task on error

jpa198:
When a task errors a new task is made instead of increasing the retry count


scripts/cybershake/auto_submit.py
```diff
@@ -152,16 +153,6 @@ def update_tasks(
                 error="Disappeared from squeue. Reset to created.",
```
jasonmotha: update error message

scripts/management/update_mgmt_db.py
```diff
@@ -1,119 +0,0 @@
...
-A script that updates a slurm mgmt db and updates the status of a task
-"""
-
-import argparse
-import qcore.constants
-from scripts.management import db_helper
-import qcore.constants as const
-
-
-def update_db(db, process, status, job=None, run_name=None, error=None):
```
jasonmotha: I think we need to restore a minimal version of this file

scripts/management/MgmtDB.py
```diff
@@ -160,22 +165,22 @@ def _check_dependancy_met(self, task, logger=workflow_logger.get_basic_logger())
...
-            (entry.status, entry.job_id, entry.retries),
+            (self.col_status, self.col_job_id),
+            (entry.status, entry.job_id),
         ):
             if value is not None:
                 cur.execute(
                     "UPDATE state SET {} = ?, last_modified = strftime('%s','now') "
-                    "WHERE run_name = ? AND proc_type = ?".format(field),
-                    (value, entry.run_name, entry.proc_type),
+                    "WHERE run_name = ? AND proc_type = ? and status != ?".format(field),
```
jasonmotha: Change to less than. So statuses can only progress forward

e2e_tests/E2ETests.py
```diff
@@ -440,6 +441,7 @@ def cancel_running(self, proc_types: List[const.ProcessType]):
                     "Checkpoint testing: Cancelling job-id {} "
                     "for {} and process type {}".format(entry[1], entry[2], entry[0])
                 )
+                self.canceled_running.append(str(entry[1]))
```
claudio525: Why as str and not int?

scripts/management/update_mgmt_db.py
claudio525: Do we use this anywhere in our workflow code? Or only for manual updates?

e2e_tests/E2ETests.py
jpa198: They get put in an sql query later on, and I think you have to use strings for that

scripts/management/update_mgmt_db.py
jpa198: This is only used for manual updates. The alternative method is to run add_to_mgmt_queue while queue_monitor is running. 
This should also have a diff for the file I re added.

e2e_tests/E2ETests.py
jasonmotha: Prepared queries (using ?) convert to str for you.

jpa198:
I ran an end to end test with a raised cancel rate and retry limit and everything worked.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@ Slurm Ground Motion Workflow
 # Changelog
 (Based on https://wiki.canterbury.ac.nz/download/attachments/58458136/CodeVersioning_v18p2.pdf?version=1&modificationDate=1519269238437&api=v2 )
 
+## [19.6.1] - 2019-06-05 -- Error task changes
+### Added
+    - A wrapper for auto_submit and queue_monitor is now available 
+    - LF and HF may now be converted to BB without the over, and IM calculations subsequently performed on them
```
jasonmotha: other

jpa198: England much hard

PR #208, opened 2019-06-19
More e2e tests

jpa198:


scripts/bb_sim.py
```diff
@@ -337,6 +337,13 @@ def unfinished():
             args.flo,
             "lowpass",
         )
+        if is_master and i == 0:
+            if len(hf_acc[:, c]) != len(lf_acc[:, c]):
+                logger.critical(
```
jasonmotha: @shot0829 How do we handle these issues? Just log and nothing to stderr?

PR #212, opened 2019-06-21
Add dt to simulation duration check

jpa198:


shared_workflow/install_shared.py
```diff
@@ -158,6 +158,14 @@ def install_simulation(
     if stat_file_path is not None:
         sim_params_dict[SimParams.stat_file.value] = stat_file_path
 
+    if float(sim_duration)/root_params_dict['dt'] - int(float(sim_duration)/root_params_dict['dt']) > 0.001:
```
jasonmotha: np is close?

PR #219, opened 2019-06-28
Changed surrounding logic so that only tasks with less failures than the

jpa198:
limit are given a new task instance only if they don't already have one
partially completed
Fixed sql


jpa198:
Tested on a cybershake database I moved to my local machine.

PR #225, opened 2019-07-03
Refactor install

jpa198:


shared_workflow/install_shared.py
```diff
@@ -215,7 +270,57 @@ def install_simulation(
...
+    root_params_base_dict, root_folder, extended_period, seed, stat_file_path, version
+):
+    root_params_base_dict["mgmt_db_location"] = root_folder
+    root_params_base_dict[RootParams.extended_period.value] = extended_period
+    root_params_base_dict[RootParams.version.value] = version
+    root_params_base_dict[RootParams.stat_file.value] = stat_file_path
+    root_params_base_dict[RootParams.stat_vs_est.value] = stat_file_path.replace(
+        ".ll", ".vs30"
+    )
+    root_params_base_dict[RootParams.stat_vs_ref.value] = stat_file_path.replace(
```
jasonmotha: @shot0829 Do we need this one?

scripts/cybershake/install_cybershake.py
```diff
@@ -108,16 +115,17 @@ def main():
...
-            fault,
-            count,
-            path_cybershake,
-            args.version,
-            args.stat_file_path,
-            args.seed,
-            args.extended_period,
-            workflow_logger.get_realisation_logger(logger, fault),
-        )
+        for i in range(1, count + 1):
```
jasonmotha: Comment to explain magic number

shared_workflow/install_shared.py
shot0829: Its not being used anymore. And there is a way to emulate the same feature by specifying a value of lfvsref under bb. So i would say its good to be removed.

jpa198:
Yeah, the refactor is in no mans land where installation is set up to be in the workflow, but isn't yet, so nothing will work

PR #227, opened 2019-07-04
added default task_config.yaml for run_cybershake

ttgenius:


scripts/cybershake/run_cybershake.py
```diff
@@ -243,6 +243,8 @@ def main():
     parser.add_argument(
         "config_file",
         help="The location of the config file containing everything to be run",
+        nargs="?",
+        default="task_config.yaml",
```
jpa198: This will look for task_config.yaml in the current working directory (Where the script was called from) rather than the directory the file is in.
Check management/install_mgmt_db for an example of how to reference a file in the same directory.

scripts/cybershake/task_config.yaml
```diff
@@ -0,0 +1,5 @@
+EMOD3D: ALL
+HF: ALL
+BB: ALL
+IM_calc: ALL
+clean_up: ALL
```
jpa198: As people are probably going to use this a the base for their own config, the other tasks (get the str_value from each task in qcore.constants.ProcessType) should be listed with the keyword NONE.

scripts/cybershake/run_cybershake.py
```diff
@@ -243,6 +243,8 @@ def main():
     parser.add_argument(
         "config_file",
         help="The location of the config file containing everything to be run",
+        nargs="?",
+        default=join(dirname(realpath(__file__)), "task_config.yaml"),
```
jpa198: The file should probably be kept in the templates directory, cybershake is getting a bit cluttered.

PR #230, opened 2019-07-10
fixed 0 usage for long usernames

ttgenius:
dashboard display 0 core hours usage for long username like melody.zhu. This is because melody.zhu is collapsed into melody.z+ and stored as melody.z+ in database.

### DIFF
--- a/dashboard/run_dashboard_app.py
+++ b/dashboard/run_dashboard_app.py
@@ -407,7 +407,7 @@ def get_total_user_chours(
         name, total_chours = app.db.get_total_user_chours(
             hpc, username, start_date, end_date
         )
-        data.append({"username": name, "total_core_hours": total_chours})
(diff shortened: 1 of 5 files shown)


dashboard/run_dashboard_app.py
```diff
@@ -407,7 +407,7 @@ def get_total_user_chours(
         name, total_chours = app.db.get_total_user_chours(
             hpc, username, start_date, end_date
         )
-        data.append({"username": name, "total_core_hours": total_chours})
+        data.append({"username": USERS[name], "total_core_hours": total_chours})
```
jasonmotha: Oh, I meant the command run on HPC to display the full name. Is that possible.

There should be options to sreport which does that.

ttgenius: reformatted sreport

PR #231, opened 2019-07-11
Checkpoint changes

jpa198:
#### Description
Stopped EMOD3D doing partial result dumps as it resulted in restart errors
Updated readme and changelog
Did catch up entry for changelog
#### Checklist
- Have you updated the README yes

### DIFF
--- a/scripts/set_runparams.py
+++ b/scripts/set_runparams.py
@@ -91,8 +91,7 @@ def create_run_params(
 
         e3d_dict["restartdir"] = os.path.join(params.sim_dir, "LF", "Restart")
         if steps_per_checkpoint:
-            e3d_dict["restart_itinc"] = int(steps_per_checkpoint)
-            e3d_dict["dump_itinc"] = int(steps_per_checkpoint)
+            e3d_dict["dump_itinc"] = e3d_dict["restart_itinc"] = int(steps_per_checkpoint)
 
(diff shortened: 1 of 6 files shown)


scripts/set_runparams.py
```diff
@@ -92,6 +92,7 @@ def create_run_params(
         e3d_dict["restartdir"] = os.path.join(params.sim_dir, "LF", "Restart")
         if steps_per_checkpoint:
             e3d_dict["restart_itinc"] = int(steps_per_checkpoint)
+            e3d_dict["dump_itinc"] = int(steps_per_checkpoint)
```
jasonmotha: Tidy up as discussed.

shared_workflow/install_shared.py
```diff
@@ -160,7 +160,7 @@ def install_simulation(
         sim_params_dict[SimParams.stat_file.value] = stat_file_path
 
     nt = float(sim_duration) / root_params_dict['dt']
-    if not isclose(nt, int(nt)):
+    if not isclose(nt, round(nt)):
```
claudio525: Why round instead of int? Shouldn't nt always be an int?

jpa198: Because if float math approximates slightly below the correct value, then using int causes it to be rounded down to the next int, instead of to the nearest correct one

PR #239, opened 2019-07-24
added black to travis

ttgenius:
Added black format checking to travis.yml


ttgenius:
> In the test log there are a significant number of files failing the test but the test appears to pass?
Travis will exit upon failing black now

ttgenius:
Thanks!

PR #248, opened 2019-08-11
fix im_calc compeletion check

ttgenius:
fixed im_calc completion check.
The reason that im_calc is shown "completed" in mgmt_db even there's runtime error is due to the missing of "not" in im_calc template when checking output file existence.

### DIFF
--- a/templates/sim_im_calc.sl.template
+++ b/templates/sim_im_calc.sl.template
@@ -35,7 +35,7 @@ fi
 
 # Check that the result files exist
 res=0
-if [[ -f {{sim_dir}}/IM_calc/{{sim_name}}.csv ]] && [[ -f {{sim_dir}}/IM_calc/{{sim_name}}_im_calc.info ]]; then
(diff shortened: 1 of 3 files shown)


templates/sim_im_calc.sl.template
```diff
@@ -35,7 +35,7 @@ fi
 
 # Check that the result files exist
 res=0
-if [[ -f {{sim_dir}}/IM_calc/{{sim_name}}.csv ]] && [[ -f {{sim_dir}}/IM_calc/{{sim_name}}_im_calc.info ]]; then
+if [[ ! -f {{sim_dir}}/IM_calc/{{sim_name}}.csv ]] && [[ ! -f {{sim_dir}}/IM_calc/{{sim_name}}_im_calc.info ]]; then
```
shot0829: the naming is incorrect.
And I think its a good time to synchronize the expected output name 
(centralize the default output file name, check IM repo)

PR #250, opened 2019-08-16
Hf sim mpi fix

sungeunbae:
#### Description
The station list is split into N blocks for N processors, where each block consists of non-consecutive stations (stride N), which ensures more evenly distributed load balancing. reducing core hours wastage. 

#### Dependencies
None. E2E test passed, and confirmed wastage reduced from 9 core hours to 1.
#### Checklist
- Have you updated the README Yes
- Have you updated the CHANGELOG  Yes


jasonmotha:
Will you look into why the auto test failed? (You'll have to run black over HF_sim.py)

sungeunbae:
> Will you look into why the auto test failed? (You'll have to run black over HF_sim.py)

Blacked

PR #251, opened 2019-08-16
Update hf_sim.py

jpa198:
Fix bugs induced by recent changes


jpa198:
Resolves issue when not all threads receive a station to work on

PR #252, opened 2019-08-19
Metadata

ttgenius:
Added metadata logging for failed runs in queue monitor.
Added attribute: run_time & err_log_time


scripts/cybershake/queue_monitor.py
```diff
@@ -151,6 +153,49 @@ def update_tasks(
                     "Disappeared from squeue. Creating a new task.",
                 )
             )
+            # When job failed, we want to log metadata as well
+            # Assumes that when enter a jobid, only one row will be will returnerd
```
shot0829: Do no "assume" while it can be forced.
sacct -X will force it not show info for steps
example:
"sacct -X -P -n -j {} -u {} -S {} -E {} -o {}".format(job_id,username,start_date,end_date,sacct_format)

scripts/cybershake/queue_monitor.py
```diff
@@ -151,6 +153,49 @@ def update_tasks(
                     "Disappeared from squeue. Creating a new task.",
                 )
             )
+            # When job failed, we want to log metadata as well
```
jpa198: Should probably be it's own function?
Could eventually be used for recording both failed and completed jobs

jasonmotha: Extract the logging metadata into a separate function as we will eventually put this logic into completed cases too.

scripts/cybershake/queue_monitor.py
```diff
@@ -151,6 +153,49 @@ def update_tasks(
...
+                        db_running_task.job_id
+                    ),
+                    shell=True,
+                )
+                .decode("utf-8")
+                .strip()
+                .split()[0]
+            )
+            # if run_time exceeds one day
+            if "-" in run_time:
```
shot0829: forcing the output to use "CPUTimeRAW"
can force the output to be in seconds, which can then avoid all these manual transformation of datetime format.

scripts/cybershake/queue_monitor.py
shot0829: yes, as we discussed in the meeting, its probably a good idea to remove the metadata logging function from the *.sl and only monitor will log them.
But no matter what way to are going to use, the functionality to query and log metadata needs to be modularized so it can be called wither within code or outside ( as @jasonmotha mentioned in his comment)

metadata/agg_json_data.py
```diff
@@ -144,18 +156,21 @@ def create_dataframe(json_files: List[str], n_procs: int, calc_core_hours: bool)
...
         if df is None:
+            print("df is none")
             df = pd.DataFrame(
                 index=[sim_name],
                 columns=pd.MultiIndex.from_tuples(columns),
                 data=np.asarray(data, dtype=object).reshape(1, -1),
             )
         else:
             # Check/Add missing columns
+            print("else")
```
shot0829: else?

scripts/cybershake/queue_monitor.py
```diff
@@ -58,11 +59,43 @@ def get_queue_entry(
...
+        .strip()
+        .split()
+    )
+    # ['578928', 'u-bl689.atmos_main.18621001T0000Z', '2019-08-16T13:05:06', '2019-08-16T13:12:56', '2019-08-16T14:58:28', '1840', '11650880', 'COMPLETED', 'nid00[166-171,180-196]']
+    submit_time, start_time, end_time = [x.replace("T", "_") for x in output[2:5]]
+    n_cores = float(output[5])
+    run_time = float(output[6]) / n_cores
+    sim_dir = sim_struct.get_sim_dir(root_folder, db_running_task.run_name)
+    log_file = os.path.join(sim_dir, "ch_log", const.METADATA_LOG_FILENAME)
+    with open("/home/melody.zhu/meta.txt", "a") as f:
```
shot0829: this looks like debug message as well? why not just use the logger? and why is it hard coded?

ttgenius: Thanks! All deleted

PR #253, opened 2019-08-21
Changed get_queued_tasks to raise an exception if a valid response is

jpa198:
not received from squeue.
Jobs are not considered to have been cancelled/WCT killed if squeue does.
not respond correctly.
Changed add_to_mgmt_queue to take in the job id if possible to ensure the correct entry is updated.
Slurm scripts have been set to provide the job id.
If the job id is not provided the update is still applied, as each update should ony apply to one entry.
If multiple rows are modified by an update the user is alerted to this.

#### Checklist

### DIFF
--- a/scripts/management/MgmtDB.py
+++ b/scripts/management/MgmtDB.py
@@ -284,7 +284,7 @@ class MgmtDB:
         """Updates all fields that have a value for the specific entry"""
         if entry.status == const.Status.queued.value:
             logger.debug(
-                "Got entry {} with status created. Setting status and job id in the db".format(
+                "Got entry {} with status queued. Setting status and job id in the db".format(
                     entry
                 )
             )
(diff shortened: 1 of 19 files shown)


scripts/management/MgmtDB.py
```diff
@@ -278,17 +278,54 @@ def _check_dependancy_met(self, task, logger=get_basic_logger()):
...
-            (self.col_job_id, self.col_status), (entry.job_id, entry.status)
-        ):
-            if value is not None:
-                cur.execute(
-                    "UPDATE state SET {} = ?, last_modified = strftime('%s','now') "
-                    "WHERE run_name = ? AND proc_type = ? and status < ?".format(field),
-                    (value, entry.run_name, entry.proc_type, entry.status),
+        if entry.status == const.Status.queued.value:
+            logger.debug(
+                "Got entry {} with status created. Setting status and job id in the db".format(
```
shot0829: status with 'queued'? instead of created

PR #254, opened 2019-08-26
specify account name for sreport user core hours

ttgenius:
#### Checklist

### DIFF
--- a/dashboard/run_data_collection.py
+++ b/dashboard/run_data_collection.py
@@ -412,7 +412,7 @@ if __name__ == "__main__":
         "--project_id",
         type=str,
         help="Specify the project ID to collect old daily core hours usages for (back from today),"
-             " default is {}".format(const.DEFAULT_ACCOUNT),
+        " default is {}".format(const.DEFAULT_ACCOUNT),
         default=const.DEFAULT_ACCOUNT,
     )
 
--- a/dashboard/run_data_collection.py
+++ b/dashboard/run_data_collection.py
@@ -195,8 +195,8 @@ class DataCollector:
 
         # user daily core hour usage
         user_ch_output = self.run_cmd(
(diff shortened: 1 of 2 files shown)


dashboard/run_old_data_collection.py
```diff
@@ -47,7 +47,7 @@ def collect_old_data(
     """Collect old core hours usage for a specified users in a previous day"""
     local_date, start_time, end_time = get_old_utc_times(day_shift)
 
-    cmd = "sreport -M {} -t Hours cluster AccountUtilizationByUser Users={} start={} end={} -n format=Cluster,Accounts,Login%30,Proper,Used".format(
+    cmd = "sreport -M {} -t Hours cluster AccountUtilizationByUser Accounts=nesi00213 Users={} start={} end={} -n format=Cluster,Accounts,Login%30,Proper,Used".format(
```
shot0829: same issue.

dashboard/run_data_collection.py
```diff
@@ -195,8 +195,8 @@ def collect_data(self, hpc: const.HPC, users: Iterable[str]):
 
         # user daily core hour usage
         user_ch_output = self.run_cmd(
-            "sreport -M {} -t Hours cluster AccountUtilizationByUser Users={} start={} end={} -n format=Cluster,Account,Login%30,Proper,Used".format(
-                hpc.value, " ".join(users), start_time, end_time
+            "sreport -M {} -t Hours cluster AccountUtilizationByUser Accounts={} Users={} start={} end={} -n format=Cluster,Account,Login%30,Proper,Used".format(
+                hpc.value, PROJECT_ID, " ".join(users), start_time, end_time
```
shot0829: 1. make 'account/project_id a parsed arg for the whole function (because soon we will have 2 projects running)
2. if not supplied, use DEFAULT_ACCOUNT from qcore.const ( instead of a local variable)

ttgenius: Done

PR #258, opened 2019-09-18
Sbatch

ttgenius:
Added sbatch script and sample core_hours.json file so that we submit/reject the job based on available core hour allocations

python sbatch.py run_bb.sl

### DIFF
--- a/scripts/sbatch.py
+++ b/scripts/sbatch.py
@@ -8,7 +8,6 @@ import argparse
 import getpass
 from enum import Enum
 
-TEST_CMD = 'mkdir test_sbatch'
 MB_PER_GB = 1024.
 
 DEFAULT_NTASKS = 1
@@ -16,6 +15,8 @@ DEFAULT_CPUS_PER_TASK = 1
 
 MAHUIKA_SLURM_CONF = '/scale_wlg_persistent/filesets/home/slurm/mahuika/etc/opt/slurm/slurm.conf'
 MAUI_SLUM_CONF = '/etc/opt/slurm.conf'
+
+SBATCH_BIN = '/opt/slurm/18.08.7/bin/sbatch'
 
 
 # can use from qcore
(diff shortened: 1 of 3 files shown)


scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+def compare_hours_and_sbatch(requested_hours, available_hours, hpc, sl_to_sbatch):
+    """
+       Compare requested hour with available hours
+       and either submit the job or reject
+    """
+    if requested_hours <= available_hours:
+        cmd = "{} -M {} {}".format(SBATCH_BIN, hpc, sl_to_sbatch)
+        subprocess.call(cmd, shell=True)
+    else:
+        sys.exit("Not enought core hours left, please contact Jonney")
```
jasonmotha: enough*

jasonmotha: Please apply for more.

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def process_slurm_header(sl_file):
+    """Read a slurm file and returns the wanted header values as a dictionary"""
+    with open(sl_file) as f:
+        lines = f.readlines()
+    # SBATCH --account=nesi00213
+    header_dict = {h.value: None for h in SlurmHeader}
+    for header in header_dict.keys():
+        for line in lines:
```
jasonmotha: Iterates over the entire slurm file {header} times. Better off doing the iteration the other way around?

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+def process_mem_per_cpu(mem_per_cpu):
+    """Convert a mem_per_cpu string to float"""
+    if isinstance(mem_per_cpu, str):
+        if mem_per_cpu[-1].upper() == "G":  # 0.5G to 0.5
+            mem_per_cpu = float(mem_per_cpu[:-1])  # 521MB to 0.5
+        elif mem_per_cpu[-2:] == "MB":
+            mem_per_cpu = float(mem_per_cpu[:-2]) / MB_PER_GB
+    else:
+        sys.exit("undefined mem per cpu, please add unit G/MB if not included")
```
jasonmotha: Doesn't it have a default unit? Or does slurm require a unit?

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("slurm_to_sbatch", help="path to slurm file to be sbatched")
+    parser.add_argument(
+        "--core_hour_json",
+        default="core_hours.json",
+        help="path to core hour json file getting from dashboard.db, default is core_hours.json",
+    )
+    parser.add_argument(
```
jasonmotha: Unsure the best way to get username. I thought this was meant to be transparent to the user?

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+        "--user", default=getpass.getuser(), help="the user who's sbatching the job"
+    )
+    args = parser.parse_args()
```
jasonmotha: Needs a catch all to get all other sbatch arguements.

scripts/sbatch.py
shot0829: Slurm does have default mem_per_cpu. its defined in the conf.
If not stated, its defualtly in MB.

shot0829: so this parameter shouldn't be 'required' as it has a default value set in the slurm.conf

scripts/sbatch.py
shot0829: a user should not be able to submit jobs under other's name (generally speaking).
so a default value set as the user who executed this script should be enough

scripts/sbatch.py
shot0829: try adding a positional argument with nargs='*'

scripts/sbatch.py
shot0829: Slurm assumes its MB

scripts/sbatch.py
shot0829: Another thing to note, is that once we added the ability to parse additional arguments for slurm itself ( as stated in another comment), we will need to check if these parameters are supplied in the args, instead of a header in *.sl

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+MB_PER_GB = 1024.0
+
+DEFAULT_NTASKS = 1
+DEFAULT_CPUS_PER_TASK = 1
+
+MAHUIKA_SLURM_CONF = (
+    "/scale_wlg_persistent/filesets/home/slurm/mahuika/etc/opt/slurm/slurm.conf"
+)
+MAUI_SLUM_CONF = "/etc/opt/slurm.conf"
+
```
shot0829: I dont think we need to differentiate slurm conf for maui/mahuika.
as the slurm we currently have is compiled to read a default location, and symbolic linked to different file based on machines, set both to "/etc/opt/slurm/slurm.conf" should be enough.
( running "realpath /etc/opt/slurm/slurm.conf" on both machine will show you the actual file each machine is reading)

jasonmotha:
https://github.com/ucgmsim/slurm_gm_workflow/pull/258#discussion_r325964385

I think that's overkill for the moment. Given that we are passing arguments through it is a hack to our system but if someone is subverting the system, we can see it in other ways.

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+        if line.startswith("PartitionName={}".format(partition_name)):
+            # PartitionName=prepost    Nodes=wbl[001-005,008-011] TRESBillingWeights="CPU=1.0,Mem=0.1429G"  QOS=p_prepost2 MaxTime=03:00:00 PriorityTier=1
+            try:
+                return get_billing_weights(line)
+            except (IndexError, ValueError):
+                # TRESBillingWeights not found, use default cpu_weights and mem_weights
+                for l in lines[::-1]:
+                    if l.startswith("PartitionName=DEFAULT"):
+                        return get_billing_weights(l)
+
```
shot0829: set if BillingWeight is not even set in partition=DEFAULT, then weight_cpu=1.0 weight_mem=0
so Maui will not needed to be treated as a special case.

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+def calculate_requested_chours(
+    cpu_billing_weights,
+    mem_billing_weights,
+    mem_per_cpu,
+    ntasks=DEFAULT_NTASKS,
+    cpus_per_task=DEFAULT_CPUS_PER_TASK,
+):
+    """Calculate requested core hours for a job on mahuika"""
```
shot0829: as stated, maui shouldnt be treated as a special case.

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+
+    if not slurm_header_dict[SlurmHeader.cpus_per_task.value]:
+        slurm_header_dict[SlurmHeader.cpus_per_task.value] = DEFAULT_CPUS_PER_TASK
+    else:
+        slurm_header_dict[SlurmHeader.cpus_per_task.value] = int(
+            slurm_header_dict[SlurmHeader.cpus_per_task.value]
+        )
+
+    if hpc == HPC.mahuika.value:
```
shot0829: this if statement shouldnt be needed, if other functions are updated to not treat maui separately

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+            mem_per_cpu = float(mem_per_cpu[:-2]) / MB_PER_GB
+    else:
+        sys.exit("undefined mem per cpu, please add unit G/MB if not included")
+    return mem_per_cpu
+
+
+def process_header_values(slurm_header_dict, slurm_conf, hpc):
+    """
+       Process header dict values
+       For maui, mem_per_cpu is None as mem is not calculated against billing
```
shot0829: somewhat incorrect. meme_per_cpu should tracked as well.
but the BillingWeight of memory is 0.

scripts/sbatch.py
```diff
@@ -0,0 +1,288 @@
...
+            slurm_conf, header_dict[SlurmHeader.partition.value]
+        )
+        req_hours = calculate_requested_chours(
+            cpu_weights,
+            mem_weights,
+            header_dict[SlurmHeader.mem_per_cpu.value],
+            header_dict[SlurmHeader.ntasks.value],
+            header_dict[SlurmHeader.cpus_per_task.value],
+        )
+    elif hpc == HPC.maui.value:
```
shot0829: this if/elif statement can be removed after other functions has been updated to handle both maui/mahuika.

scripts/sbatch.py
jasonmotha: What about cross platform submission?

shot0829: totally forgot about that. in that case, we  have to hard-coded it, but have to update everytime when they decided to change the config.
There isnt a good way i can think of to get this dynamically, since each machine can only see certain file system.

scripts/sbatch.py
```diff
@@ -144,74 +152,53 @@ def get_available_chours(json_file, account, username):
...
-                value = line.strip().split("--{}=".format(header))[-1]
-                header_dict[header] = value
-                break
+    for line in lines:
+        if line.startswith("#") and "SBATCH" in line.upper():
+            for header in header_dict.keys():
+                if "{}=".format(header) in line:
+                    value = line.strip().split("--{}=".format(header))[-1]
+                    header_dict[header] = value
+                    break
```
jasonmotha: stop if parsed further than header? @shot0829

scripts/sbatch.py
```diff
@@ -224,13 +211,26 @@ def get_hpc_conf(partition):
...
     return hpc, slurm_conf
 
 
+def compare_hours_and_sbatch(requested_hours, available_hours, hpc, sl_to_sbatch, sbatch_extra_args):
+    """
+       Compare requested hour with available hours
+       and either submit the job or reject
+    """
+    if requested_hours <= available_hours:
+        cmd = "{} -M {} {} {}".format(SBATCH_BIN, hpc, sl_to_sbatch, sbatch_extra_args)
```
jasonmotha: What happens if -M was specified in this command's sbatch?

scripts/sbatch.py
```diff
@@ -246,43 +246,39 @@ def main():
...
-
+    print(header_dict)
     # Get hpc_name and path to slurm.conf
     hpc, slurm_conf = get_hpc_conf(header_dict[SlurmHeader.partition.value])
 
     # Process headers for calculation
-    header_dict = process_header_values(header_dict, slurm_conf, hpc)
-
+    header_dict = process_header_values(header_dict, slurm_conf)
+    print(header_dict)
```
jasonmotha: Debug

PR #263, opened 2019-10-07
Sbatch'

shot0829:
#### Checklist


scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+        obj.id = id
+        obj.type = type
+        obj.default = default
+        return obj
+
+
+#TODO:nest partition classes with HPC classes so no double define
+class MahuikaPartition(Enum):
+    large = (0, "large", 72, 108*MB_PER_GB)
+    long = (1, "long", 72, 108*MB_PER_GB)
```
jpa198: Should avoid using long as a variable name?

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    perpost = (2, "prepost", 72, 480*MB_PER_GB)
```
jpa198: Spelling

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    bigmem = (3, "bigmem", 72, 480*MB_PER_GB)
+    hugmem = (4, "hugemem", 128, 4000*MB_PER_GB)
```
jpa198: Also spelling

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+        # return a value from narrow to broad
+        if partition_keyword_value is not None:
+            return partition_keyword_value
+        elif default_partition_keyword_value is not None:
+            return default_partition_keyword_value
+        elif default_keyword_value is not None:
+            return default_keyword_value
+        else:
+            #last catch of errors if something unexpected happened
+            raise ValueError("keyword:{} were found but none were parsed correctly".format(keyword))
```
jpa198: "was found"

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+        if ntasks_per_node is not SlurmHeader.ntasks_per_node.default:
+            nodes = nodes * (ntasks/ntasks_per_node)
+        total_cpus = nodes * cpu_per_node
+        total_mem  = nodes * mem_per_node
+    else:
+        total_mem = total_cpus * mem_per_cpu
+    #TODO:GPU may need to be included
+    requested_hours = (total_seconds_requested * (total_cpus * cpu_billing_weights)) + (
+        total_mem * mem_billing_weights
+    )
```
jpa198: WTC is (Core seconds) + (Mem per second) ? Might need to check brackets

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    if 'G' not in mem_per_cpu.upper():  # no unit (default =MB) or unit=MB, transforming to GB
+        value = float(value) / MB_PER_GB
+    
+    return value
+
+
+def process_header_values(slurm_header_dict, slurm_conf):
+    """
+       Process header dict values
+       get defaul values if not provided
```
jpa198: Spelling

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    parser.add_argument(
+        "--user", default=getpass.getuser(), help="the user who's sbatching the job"
+    )
+    args = parser.parse_args()
+
+    assert os.path.exists(args.slurm_to_sbatch)
+    assert os.path.exists(args.core_hour_json)
+
+    # Get raw headers
+    header_dict = process_slurm_header(args.slurm_to_sbatch)
```
jpa198: Should be merged with sbatch_extra_args to check for variables. 
Probably don't need to check for environment variables, but worth considering

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    raise IndexError("no matching partition name found")
+
+
+def compare_hours_and_sbatch(requested_hours, available_hours, hpc, sl_to_sbatch, sbatch_extra_args):
+    """
+       Compare requested hour with available hours
+       and either submit the job or reject
+    """
+    if requested_hours <= available_hours:
+        cmd = "{} -M {} {} {}".format(SBATCH_BIN, hpc, sl_to_sbatch, sbatch_extra_args)
```
jpa198: " ".join(sbatch_extra_args)
Otherwise the list will be put into the string e.g. "['--ncores' '40']"

scripts/core_hours.json
```diff
@@ -0,0 +1,17 @@
+[{"nesi00213":
```
jasonmotha: sample file shouldn't really be committed?

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+        obj.id = id
+        obj.cpus = cpu_per_node
+        obj.mem = mem_per_node
+        obj.config_path = MAHUIKA_SLURM_CONF
+        obj.hpc = HPC.mahuika.value
+        return obj
+
+
+class MauiPartition(Enum):
+    nesi_research = (0, "nesi_research", 80, 160*MB_PER_GB)
```
jasonmotha: Use a named tuple or class, or split into attributes.

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+    """
+       Read a slurm.conf and returns the correspoding cpu_weights and mem_weights according to the partition_name
+       if nothing were found, a default of cpu=1.0 mem=0 will be used
+    """
+    with open(slurm_conf, "r") as f:
+        lines = f.readlines()
+    #initialize
+    TRES_dict = {vl.value: None for vl in varLevel}
+    
+    # PartitionName is at the end of file
```
jasonmotha: Umm, it shouldn't be.

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+def process_slurm_header(sl_file):
+    """Read a slurm file and returns the wanted header values as a dictionary"""
+    with open(sl_file) as f:
+        lines = f.readlines()
+    #example line: SBATCH --header=value
+    header_dict = {h.value: h.default for h in SlurmHeader}
+    for line in lines:
+        if line.upper().startswith("#SBATCH"):
```
jasonmotha: short circuit rather than checking the entire file?

scripts/sbatch.py
```diff
@@ -0,0 +1,476 @@
...
+                    #special case for WCT(time) parameter
+                    if header is SlurmHeader.time:
+                        #a day is specified, i.e. "%d-%H:%M:%S"
+                        wct_list = list( map(int, re.findall(r"[-+]?\d*\.\d+|\d+", value)))
+                        #filter the value to contain 4 fields, day, h, m, s
+                        while len(wct_list) < 4: 
+                            wct_list.insert(0,0)
+                        value = datetime.timedelta(days=wct_list[0],hours=wct_list[1],minutes=wct_list[2],seconds=wct_list[3]).total_seconds()
+                    header_dict[header.value] = header.type(value)
+                    break
```
jasonmotha: doesn't actually do anything.

scripts/sbatch.py
shot0829: this was a logic from melody's.....
I'll fix that.

shot0829: my comment even said scan all lines @.@

scripts/sbatch.py
shot0829: since the its looping through all potential header, it breaks when a match is found?

jasonmotha: Yep. That's right carry on.

jasonmotha:
Don't forget to update the changelog.

scripts/sbatch.py
shot0829: there is some automated value assign in the __new__, so if its changed to named tuple than there will be a lot of repeittive values. (i.e. hpc num and conf for each partition)

scripts/sbatch.py
shot0829: how do we determine when to stop? any non-blank-line or not-start-with-hash?

jasonmotha: Yep. That's what I'd do. I'd almost say that an SBATCH directive after the header is bad news anyway.

scripts/sbatch.py
jasonmotha: A struct/class is probably better then. (just to give the parameters some naming rather than 1, 2, 3)

scripts/sbatch.py
shot0829: this will introduce all new level of complexity.
since we dont use this behavior often, i would say its on low priority.
Its will be marked as a TODO.

jasonmotha: Agreed. I already reckon we've gone overboard with this already. We shouldn't have actually done slurm parsing as a first step.

scripts/sbatch.py
shot0829: but then iterations will need to use dictionaries instead of just a list/array?

PR #264, opened 2019-10-08
Add check to make sure flo in the vm_params and root_params match

jpa198:

### DIFF
--- a/scripts/cybershake/install_cybershake_fault.py
+++ b/scripts/cybershake/install_cybershake_fault.py
@@ -168,12 +168,17 @@ def install_fault(
             extended_period=extended_period,
         )
 
-        if not isclose(vm_params_dict['flo'], root_params_dict['flo']):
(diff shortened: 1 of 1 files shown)


scripts/cybershake/install_cybershake_fault.py
```diff
@@ -167,12 +168,25 @@ def install_fault(
...
+            root_params_dict, fault_params_dict, sim_params_dict, vm_add_params_dict = (
+                None,
+                None,
+                None,
+                None,
+            )
+
         if (
             root_params_dict is None
             and fault_params_dict is None
```
jasonmotha: or?

jasonmotha: Or don't we necesarily need all of them?

jpa198: They're only None if there's an nt mismatch in the install_simulation function, or an flo mismatch.
This is just a stop gap till the prepro/install refactor can get done/in

PR #269, opened 2019-10-29
Prevent $? bug

jpa198:
There was a bug found where failed runs would be considered successfully completed, even though they had failed. This attempts to prevent this bug. Bug found during other testing, solution developed there.

#### Checklist


jasonmotha:
$? is so dangerous. Good catch.

PR #282, opened 2019-12-09
Unperturbated empiricals

jpa198:
#### Description
Rrups has been updated, a couple of other bug fixes are in there too.
Workflow empiricals still don't work, current solution is a manual 'run once' at the end of a simulation run that has had rrups run
Empiricals are closer to working, but didn't want to spend too much more time on it when a manual solution will be available
If an event/a fault only has one realisation to be run it will be installed without the _REL extension
Run_cybershake should stop properly now, but with a longer time out
#### Checklist


templates/empirical.sl.template
```diff
@@ -10,27 +13,27 @@ echo ___calculating empirical___
...
-    time $EMP_PATH/emp_aggregation.py -o {{output_dir}} -i {{run_name}} -r {{fault_name}} {{output_dir}}/{{run_name}}_*.csv
-    time $gmsim/workflow/verification/calculate_epsilon.py {{mgmt_db_location}}/{{fault}}/IM_calc/{{run_name}}/{{run_name}}.csv {{output_dir}}/{{run_name}}.csv {{output_dir}}/{{run_name}}_epsilon.csv
+    time $EMP_PATH/empirical/scripts/calculate_empirical.py {{output_dir}} -i {{run_name}} {{extended}} -r {{rrup_file}} -srf {{srf_info}} --vs30_file /nesi/project/nesi00213/StationInfo/non_uniform_whole_nz_with_real_stations-hh400_v18p6.vs30
+    time $EMP_PATH/empirical/scripts/emp_aggregation.py {{output_dir}}/{{run_name}}_*.csv -o {{output_dir}} -i {{run_name}} -r {{fault_name}}
+    time $gmsim/workflow/verification/calculate_epsilon.py {{mgmt_db_location}}/Runs/{{fault_name}}/{{run_name}}/IM_calc/{{run_name}}.csv {{output_dir}}/{{run_name}}.csv {{output_dir}}/{{run_name}}_epsilon.csv
 
 #test before update mgmt_db -- Needs a test to see if files are output
     timestamp=`date +%Y%m%d_%H%M%S`
     res=`echo 0`
```
jasonmotha: Can update the test to check that there are more than 1 line in the output files?


Otherwise this one looks good.

jpa198: The empiricals slurm script still doesn't work, it's just that I didn't want the work I'd put into it to disappear after this PR

PR #287, opened 2019-12-17
Trim boundary stations

jpa198:
#### Description
Trim stations that have broken components, and update e2e test data to reflect this
#### Checklist


shared_workflow/install_shared.py
```diff
@@ -353,7 +353,7 @@ def generate_fd_files(
     sxy = []
     suname = []
     for i in range(len(xy)):
-        if xy[i] is None:
+        if xy[i] is None or xy[i][0] == nx-1 or xy[i][1] == ny-1:
```
shot0829: this reminds me of the issue we had a while ago, where duplicate stations were treated as first-come-first-serve, which cause some real stations removed. lets have a discussion with robin and see if its suitable to fit it in this PR.

PR #289, opened 2020-01-26
Hf version 546 update

jpa198:
#### Description
Updated versions and test data to HF 5.4.6
Also did some work on the db query tool while I was waiting for things to run/wanted to see a better breakdown while running the tests

#### Checklist
- Have you updated the CHANGELOG no


scripts/management/query_mgmt_db.py
```diff
@@ -31,22 +73,97 @@ def print_run_status(db, run_name, error=False, count=False, config_file=None):
...
-                        """,
-                    (i,),
-                ).fetchone()[0]
-            )
-        print(
-            "created: {}, queued: {}, running: {}, completed: {}, failed: {}, other: {}, total: {}".format(
-                *vals, sum(vals)
-            )
-        )
+        if detailed_count:
```
jasonmotha: This block is a behemoth. I'm guessing it could be simplified / functioned out? Also, maybe it might be a better fit in a separate script?

scripts/management/query_mgmt_db.py
```diff
@@ -9,18 +9,60 @@
 
 from shared_workflow.shared_automated_workflow import parse_config_file
 
+PATTERN_FORMATTER = "{}, {}: created: {}, queued: {}, running: {}, completed: {}, failed: {}, other: {}, total: {}"
 
-def print_run_status(db, run_name, error=False, count=False, config_file=None):
+
+def simple_query_builder(
```
jasonmotha: Calling this simple seems indescriptive? maybe just `query_builder`

scripts/management/query_mgmt_db.py
```diff
@@ -9,42 +9,56 @@
...
 
-def simple_query_builder(
+def state_table_query_builder(
     what,
-    where_from,
     state=None,
     process_type=None,
-    run_name_exact=None,
-    run_name_similar=None,
+    run_name_exact: bool = None,
```
jasonmotha: Default should be False?

PR #291, opened 2020-02-03
Remove security vulnerability

jasonmotha:
#### Checklist


install_workflow/maui_python3_requirements.txt
```diff
@@ -23,7 +23,7 @@ ipython==7.2.0
 ipython-genutils==0.2.0
 itsdangerous==1.1.0
 jedi==0.13.2
-Jinja2==2.10
+Jinja2
```
jpa198: Does this behaviour allow 2.09, or will it force an install of a newer version? May be better off with >=[current version]

PR #292, opened 2020-02-10
Default HF rayset is changed from 1 2 to 1

sungeunbae:
#### Description
Default HF rayset is changed from 1 2 to 1, which is internally handled as [1], not [1,2] as it used to be. 
#### Dependencies
None
#### Checklist
- Have you updated the README : No need
- Have you updated the CHANGELOG : Yes


sungeunbae:
@jasonmotha Merge please?

PR #298, opened 2020-03-19
reverted/added back seekbyte into the 'base' hf_sim_params for easier…

shot0829:
… track. also added back a line of missing params from last merge

#### Checklist

### DIFF
--- a/scripts/hf_sim.py
+++ b/scripts/hf_sim.py
@@ -14,7 +14,7 @@ import numpy as np
 import logging
 
 from shared_workflow import shared_defaults
-from qcore import binary_version, MPIFileHandler, constants
+from qcore import binary_version, MPIFileHandler, constants, utils
 
 comm = MPI.COMM_WORLD
 rank = comm.Get_rank()
--- a/scripts/hf_sim.py
+++ b/scripts/hf_sim.py
@@ -406,12 +406,13 @@ def run_hf(local_statfile, n_stat, idx_0, velocity_model=args.velocity_model):
         "",
     ]
     if utils.compare_versions(args.version, "6.0.3") >= 0:
-        hf_sim_args.insert(-2,
+        hf_sim_args.insert(
+            -2,
             "{} {} {}".format(
                 args.stress_param_adj[0],
                 args.stress_param_adj[1],
                 args.stress_param_adj[2],
-            )
+            ),
         )
 
     stdin = "\n".join(hf_sim_args)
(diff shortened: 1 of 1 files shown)


scripts/hf_sim.py
```diff
@@ -400,19 +400,20 @@ def run_hf(local_statfile, n_stat, idx_0, velocity_model=args.velocity_model):
         str(args.vs_moho),
         "%d %s %s %s %s %d" % (nl_skip, vp_sig, vsh_sig, rho_sig, qs_sig, ic_flag),
         velocity_name,
+        "%s %s %s" % (args.fa_sig1, args.fa_sig2, args.rv_sig1),
```
jasonmotha: Doesn't this screw up for previous versions of HF eg. 5.4.5.4

shot0829: well, this is actually what the previous version needed, this line was somehow skipped from the last PR #294 
took me a while to find what was missing.
And this is why i want to keep the original 'template', so in the future if more params were needed, the base will not change.

scripts/hf_sim.py
```diff
@@ -402,18 +402,22 @@ def run_hf(local_statfile, n_stat, idx_0, velocity_model=args.velocity_model):
         velocity_name,
         "%s %s %s" % (args.fa_sig1, args.fa_sig2, args.rv_sig1),
         str(args.path_dur),
-        str(head_total + idx_0 * (nt * N_COMP * FLOAT_SIZE)),
         "",
     ]
+    
+    # extra params needed for v6.0
     if utils.compare_versions(args.version, "6.0.3") >= 0:
         hf_sim_args.insert(
```
jasonmotha: Better to change this back to append?

scripts/hf_sim.py
```diff
@@ -402,18 +402,22 @@ def run_hf(local_statfile, n_stat, idx_0, velocity_model=args.velocity_model):
...
+            -1,
             "{} {} {}".format(
                 args.stress_param_adj[0],
                 args.stress_param_adj[1],
                 args.stress_param_adj[2],
             ),
         )
+    # add seekbyte for qcore adjusted version
+    if bin_mod:
+        hf_sim_args.insert(-1, str(head_total + idx_0 * (nt * N_COMP * FLOAT_SIZE)))
```
jasonmotha: Same here?

scripts/hf_sim.py
shot0829: it needs the "' at the end of the list, for a extra newline. 
so unless we removed it from template and then manually add in '\n', we need to insert to the -1 position

jasonmotha: Ah yeah. Lets just append the "" at the end of the argument building. It makes it less bug prone.

PR #299, opened 2020-03-30
made argparese in hf and bb to be importable

shot0829:
#### Checklist

### DIFF
--- a/scripts/cybershake/install_cybershake_fault.py
+++ b/scripts/cybershake/install_cybershake_fault.py
@@ -274,7 +274,6 @@ def install_fault(
 
         command_template, add_args = bb_gen_command_template(sim_params)
         run_command = gen_args_cmd(ProcessType.BB, sim_params.sim_dir, ProcessType.BB.command_template, command_template, add_args)
-        print(run_command)
         bb_args_parser(cmd=run_command)
         # change back, to prevent unexpected error
         sys.argv[0] = main_script_name
--- a/scripts/cybershake/install_cybershake_fault.py
+++ b/scripts/cybershake/install_cybershake_fault.py
@@ -84,7 +84,6 @@ def gen_args_cmd(
     command_parts = command_template.format(**template_parameters).split()
     #remove srun, python, and *.py from the command
     for i in list(command_parts):
-        print(f'i: {i}')
         if (any([ i == x for x in ['srun','python']]) or i[-3:] == '.py'):
             command_parts.remove(i)
             
--- a/scripts/cybershake/install_cybershake_fault.py
+++ b/scripts/cybershake/install_cybershake_fault.py
@@ -5,13 +5,14 @@ Called from inside a loop in install_cybershake.sh
 """
 
 import os
+import sys
 import glob
 import argparse
 from logging import Logger
 
-#from scripts.bb_sim import args_parser as bb_args_parser
+from scripts.bb_sim import args_parser as bb_args_parser
 from scripts.hf_sim import args_parser as hf_args_parser
-#from scripts.submit_bb import gen_commmand_template as bb_gen_command_template
+from scripts.submit_bb import gen_command_template as bb_gen_command_template
 from scripts.submit_hf import gen_command_template as hf_gen_command_template
 
 from numpy import isclose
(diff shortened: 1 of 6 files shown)


shot0829:
DO NOT MERGE!!
this PR is to show the difference of style of update.
i have not commit other codes needed.

scripts/hf_sim.py
```diff
@@ -163,63 +169,132 @@ def random_seed():
         default=["1", "-1", "-1"],
     )
 
-    try:
-        args = parser.parse_args()
-    except SystemExit as e:
-        print(e, flush=True)
-        # invalid arguments or -h
-        comm.Abort()
+    args = parser.parse_args(cmd)
```
jpa198, line 172: Clever, leaving it as None results in default behaviour. Maybe add that to the docstring though?

jpa198:
The diff is mostly showing the same content moved around, are there any differences aside from the top level content being in main function for each of them?

shot0829:
> The diff is mostly showing the same content moved around, are there any differences aside from the top level content being in main function for each of them?

modulizing those functions means that it a variable has to passed around, when there is no forced class/definition of the structure of that variable, which may lead to future confusions.

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -64,6 +72,31 @@ def main():
...
+        if (any([ i == x for x in ['srun','python']]) or i[-3:] == '.py'):
+            command_parts.remove(i)
+            
+    for key in add_args:
+        command_parts.append("--" + key)
+        if add_args[key] is True:
+            continue
+        command_parts.append(str(add_args[key]))
+
+    return (list(map(str, command_parts)))
```
jpa198: All elements of command_parts seem to be a string already?

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -223,6 +253,41 @@ def install_fault(
             vm_params_dict,
         )
 
+        # test if the params are accepted by steps HF and BB
+        sim_params = utils.load_sim_params(os.path.join(sim_dir, "sim_params.yaml"))
+        # check hf
+
+        # temporary change the script name to hf_sim, due to how error message are shown
+        main_script_name = sys.argv[0]
+        sys.argv[0] = "hf_simi.py"
```
jasonmotha, line 262: hf_sim.py ??

scripts/submit_bb.py
```diff
@@ -18,6 +18,19 @@
 default_wct = "00:30:00"
 
 
+def gen_command_template(params):
+    command_template_parameters = {
+        "outbin_dir": sim_struct.get_lf_outbin_dir(params.sim_dir),
```
jasonmotha, line 23: consistency. Probably best to use simulation_structure everywhere.

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -64,6 +73,27 @@ def main():
...
+def gen_args_cmd(
+    process: ProcessType, sim_dir, command_template, template_parameters, add_args={}
+):
+    # this function is adapted from generate_command from shared_workflow.shared_template
+    command_parts = []
+
+    command_parts = command_template.format(**template_parameters).split()
+    # remove srun, python, and *.py from the command
+    for i in list(command_parts):
+        if any([i == x for x in ["srun", "python"]]) or i[-3:] == ".py":
```
jpa198, line 85: `if i in ["srun", "python"] or i.endswith(".py")`?

scripts/cybershake/install_cybershake_fault.py
shot0829: oops

scripts/cybershake/install_cybershake_fault.py
jpa198: Is there a reason for the list comprehension still being in there?

scripts/cybershake/install_cybershake_fault.py
shot0829: I think its better to keep it there, as it is not computational heavy (can only be so much args) and work as a last safety check if any unexpected values are in there (very unlikely but better be safe than sorry).

PR #303, opened 2020-04-07
Slack alert and hf scp

shot0829:
#### Checklist

### DIFF
--- a/scripts/cybershake/add_to_mgmt_queue.py
+++ b/scripts/cybershake/add_to_mgmt_queue.py
@@ -33,6 +33,7 @@ if __name__ == "__main__":
         "--error",
         type=str,
         help="Errors that occurred during the execution of the script.",
+        default=None,
     )
 
     args = parser.parse_args()
--- a/scripts/cybershake/queue_monitor.py
+++ b/scripts/cybershake/queue_monitor.py
@@ -209,6 +209,7 @@ def queue_monitor_loop(
     sleep_time: int,
     max_retries: int,
     queue_logger: Logger = qclogging.get_basic_logger(),
+    alert_url=None,
 ):
     mgmt_db = MgmtDB(sim_struct.get_mgmt_db(root_folder))
     queue_folder = sim_struct.get_mgmt_db_queue(root_folder)
@@ -300,7 +301,7 @@ def queue_monitor_loop(
 
         if len(entries) > 0:
             queue_logger.info("Updating {} mgmt db tasks.".format(len(entries)))
-            if mgmt_db.update_entries_live(entries, max_retries, queue_logger):
+            if mgmt_db.update_entries_live(entries, max_retries, queue_logger, alert_url=alert_url):
                 for file_name in entry_files:
                     os.remove(os.path.join(queue_folder, file_name))
             else:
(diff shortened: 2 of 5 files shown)


scripts/cybershake/add_to_mgmt_queue.py
```diff
@@ -33,6 +33,7 @@
         "--error",
         type=str,
         help="Errors that occurred during the execution of the script.",
+        default=None,
```
jpa198, line 36: Optional arguments are None by default?

scripts/management/MgmtDB.py
```diff
@@ -94,6 +96,14 @@ def update_entries_live(
...
                     self._insert_task(cur, realisation_name, process)
                     logger.debug("New task added to the db")
+                    #send light warnings to alert_url
+                    if alert_url != None:
+                        msg = f"fault:{entry.run_name} step:{entry.proc_type} has failed with error:{entry.error}"
+                        r = requests.post(alert_url, json={'text':msg})
+                elif (entry.status == const.Status.failed.value and self.get_retries(process, realisation_name) >= retry_max) and alert_url != None:
+                    #task failed and meet the cap or retries, send a @here to alert_url for higher attention
+                    msg = f"@here fault:{entry.run_name} step:{entry.proc_type} has failed with error:{entry.error} and meet the retry cap"
+                    r = requests.post(alert_url, json={'text':msg})
```
jpa198, line 106: This should be in a separate function. Preferably in a separate file.
MgmtDB shouldn't be sending internet posts.

scripts/management/MgmtDB.py
```diff
@@ -1,6 +1,7 @@
 import os
 import sqlite3 as sql
 from logging import Logger
+import requests
```
jpa198, line 4: Add requests to the requirements.txt file

scripts/cybershake/add_to_mgmt_queue.py
shot0829: yeah, this shouldnt effect anything, it is just for debugging purposes. ( was having some issue when running on old stuffs)

scripts/management/MgmtDB.py
jpa198: Oh god, you actually have to add requests to the docker library...

scripts/management/MgmtDB.py
shot0829: good point. i can move this up an level to queue_monitor

scripts/cybershake/queue_monitor.py
```diff
@@ -29,6 +30,18 @@
...
+def send_alert(msg, alert_url):
+    data = {"text": msg}
+
+    req = urllib.request.Request(alert_url)
+    # create header and data, slack-bot requires json data
+    req.add_header("Content-Type", "application/json; charset=utf-8")
+    jsondata = json.dumps(data)
+    jsondataasbytes = jsondata.encode("utf-8")
+    req.add_header("Content-Length", len(jsondataasbytes))
+    response = urllib.request.urlopen(req, jsondataasbytes)
```
jpa198, line 42: I think this should be moved to its own file, but it's ok for now. 
If we get any more internet communication stuff it should be moved in future.

PR #306, opened 2020-04-30
resetting the default for raysets as this influences this simulation …

jasonmotha:
…value. Changing the value in the latest simulation version instead


sungeunbae:
@jasonmotha  Run Black and push again.

PR #314, opened 2020-06-04
Multi task running

jasonmotha:
More flexibility for running tasks!!!1

#### Checklist
- Have you updated the CHANGELOG yes?

### DIFF
--- a/scripts/cybershake/run_cybershake.py
+++ b/scripts/cybershake/run_cybershake.py
@@ -226,11 +226,16 @@ def parse_config_file(
         elif pattern == NONE:
             pass
         else:
-            if pattern == ONCE:
-                pattern = ONCE_PATTERN
-            if pattern not in tasks_with_pattern_match.keys():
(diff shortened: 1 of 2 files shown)


scripts/cybershake/run_cybershake.py
```diff
@@ -226,11 +226,14 @@ def parse_config_file(
...
             pass
         else:
-            if pattern == ONCE:
-                pattern = ONCE_PATTERN
-            if pattern not in tasks_with_pattern_match.keys():
-                tasks_with_pattern_match.update({pattern: []})
-            tasks_with_pattern_match[pattern].append(proc)
+            if isinstance(pattern, str):
+                pattern = [pattern]
+            for i in pattern:
```
jpa198, line 231: i should have a better name e.g. subpattern or match

jasonmotha: I agree. I couldn't think of a good one. Subpattern is a good one.

PR #316, opened 2020-06-10
Cascading failures

jasonmotha:
So when you fail Emod (and may need to rerun it) it also marks any dependant completed tasks as failed.
So if emod, HF, BB was complete but we needed to rerun emod, marking emod as failed will fail BB so it won't try to run IM_Calc till BB is re-run. 

Files will need to be deleted separately. for corrupted runs. Better pre-state checks are a separate "feature".

### DIFF
--- a/scripts/management/MgmtDB.py
+++ b/scripts/management/MgmtDB.py
@@ -117,7 +117,7 @@ class MgmtDB:
                                         const.Status.completed.value,
                                         entry.run_name,
                                     ),
-                                ).fetchone()
+                                ).fetchone()[0]
                                 print(job_id)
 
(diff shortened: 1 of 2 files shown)


scripts/management/MgmtDB.py
```diff
@@ -117,6 +130,34 @@ def update_entries_live(
 
         return True
 
+    @staticmethod
+    def find_dependant_task(self, cur, entry):
```
jpa198, line 136: A static method with self?

jasonmotha: That's the last time I listen to IDE feedback... (fixed lol)

PR #332, opened 2020-08-21
binary tests for occurances of 0 in the output

jasonmotha:

### DIFF
--- a/scripts/test_binary.py
+++ b/scripts/test_binary.py
@@ -63,14 +63,12 @@ if __name__ == "__main__":
 
     # binary zero check
     # Checks 10 random stations for any occurances of 0 in the output (aka results have not been written)
-    for i in range(10):
-        station_idx = np.random.choice(bin.stations.shape[0])
+    for stat_name in np.random.choice(bin.stations.name, replace=False, size=min(10, bin.stations.shape[0])):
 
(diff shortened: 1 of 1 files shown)


scripts/test_binary.py
```diff
@@ -61,6 +61,22 @@
             print("Some vs == 0, {} incomplete".format(args.process_type))
         sys.exit(1)
 
+    # binary zero check
+    # Checks 10 random stations for any occurances of 0 in the output (aka results have not been written)
+    for i in range(10):
```
claudio525, line 66: possible simplification (imo) if you want:
```python
for station_name in np.random.choice(bin.stations.name, replace=False):  
    acc = bin.acc(station_name)
```

Assumes that `bin.stations.name` is a list or array

claudio525 (approved):

claudio525 (dismissed):

jpa198 (approved):
Stations will never have a 0 value for acceleration?
Also, given LFSeis, is there any reason this couldn't be applied to LF binaries? (distant stations probably have 0s before the GM frontier hits them though)

jasonmotha:
As discussed: Yep. They're small but not 0. And LF has it's own thorough suite of testing.

PR #338, opened 2020-10-21
Update activate_maui_python3_virtenv.sh

jpa198:
#### Checklist


shot0829:
My last PR moved these sequence purposely so that the module load command will not modify the PYTHONPATH.
The new line in vene should make it retreat to globale libs when its not installed in virtual env

jasonmotha:
Jonney's fix is better.

PR #341, opened 2020-11-04
new binary test

jasonmotha:
This seems to work. But E2Etests seem to be a bit random.

### DIFF
--- a/scripts/test_binary.py
+++ b/scripts/test_binary.py
@@ -70,14 +70,14 @@ if __name__ == "__main__":
     ):
 
         acc = bin.acc(stat_name)
-
-        if np.any(acc == 0):
-            if args.verbose:
-                print(
-                    f"The velocities for station {stat_name} contains zero/s, please investigate. This "
-                    f"is most likely due to crashes during HF or BB resulting in no written output."
-                )
-            #  sys.exit(1)  # Commenting this line out until further compatibility investigations - Jason: 20-10-30
+        for comp in acc.T:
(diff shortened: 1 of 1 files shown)


jasonmotha:
E2E tests were marking these as failed though they look to be correct / passing.
`/nesi/nobackup/nesi00213/RunFolder/EndToEndTest/tests/tmp_20201104_021135`

jpa198 (approved):

scripts/test_binary.py
```diff
@@ -65,19 +65,20 @@
...
         acc = bin.acc(stat_name)
-
-        if np.any(acc == 0):
-            if args.verbose:
-                print(
-                    f"The velocities for station {stat_name} contains zero/s, please investigate. This "
-                    f"is most likely due to crashes during HF or BB resulting in no written output."
-                )
-            #  sys.exit(1)  # Commenting this line out until further compatibility investigations - Jason: 20-10-30
+        for comp in acc.T:
```
shot0829, line 74: will it not be better to use transpose()? Last time I had issue with this while using newer version of numpy and pandas.

jasonmotha: Apparently they're the same operation? https://stackoverflow.com/questions/4068177/whats-the-difference-of-numpy-ndarray-t-and-numpy-ndarray-transpose-when-self

shot0829: ah, yes, you are correct. I just double checked my commit, it was the returned datatype that had issue, not the method.

shot0829 (approved):

PR #342, opened 2020-11-10
Mgmtqrylst

jasonmotha:


jpa198 (approved):
As long as this works empirically I'm happy with it

jasonmotha:
I tested it with the 4 different modes all set with --conf and they worked so I'm happy.

shot0829 (approved):

PR #350, opened 2021-01-07
Why did the thing fail the test?

jpa198:
#### Checklist


jasonmotha (approved):

jasonmotha:
@sungeunbae Are you able to have a look at why this failed?

sungeunbae:
The auto test checks the number of lines matches the benchmark slurm script. This PR adds another line to the slurm script, causing this test to fail. Altered the test so that it will only sanity-check if the slurm script to write is of non-zero length. Actual contents of the slurm script will be tested by end-to-end test.

jasonmotha (approved):

PR #352, opened 2021-01-28
Add vm perturbations to cybershake install.

jpa198:
Two new flags were added to install_cybershake.
If both are given an error is raised.
Perturbations will be used or ignored depending on the flag given.
If neither is given but perturbation files exist an error is raised.

Minor spelling and variable name changes for clarity

#### Description
Will update the run manual on the wiki as part of this

#### Dependencies
Qcore PR with the same name

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@ Slurm Ground Motion Workflow
 # Changelog
 (Based on https://wiki.canterbury.ac.nz/download/attachments/58458136/CodeVersioning_v18p2.pdf?version=1&modificationDate=1519269238437&api=v2 )
 
+## [19.6.27] - 2021-01-28 -- Added VM perturbations on install
+### Added
+    - Flags --vm_perturbations and --ignore_vm_perturbations for use with velocity model perturbations
+    - If used the sim_params and root_params files will be updated for emod3d to use the relevant perturbation file
+
 ## [19.6.26] -  2020-12-04 -- Migrated to Jenkins CI
 ### Changed
     - Removed .travis.yml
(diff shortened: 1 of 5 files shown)


shot0829 (dismissed):

jasonmotha (approved):

jasonmotha:
I'm pretty sure FileExistsError isn't meant to be used like that but I guess it kinda makes sense.

claudio525 (commented):
Why use two flags, can't the same behavior be achieved with one? I.e just have flag vm_perturbations, which defaults to False if not set. This still covers all 3 scenarios?
1) Files and flags given -> use perturbations
2) No files and no flag given -> no perturbations
3) Files exists but no flag given -> raise error

jasonmotha:
The reason for two flags is it helps prevent unintended behavior

1) (current) no perturbations, no flags
2) perturbations exist, no flags -> error
3) perturbations exist, pert_flag -> okay
4) perturbations exist, no_pert_flag -> okay (but user is forced to recognise there are perturbations and ignoring them)
5) no perturbations, pert_flag -> error

Otherwise, if you forget the flag, you'll get an installed version without perturbations and you are unlikely to realise that. We want installation errors to be caught at run-time.

claudio525:
Hmh I suppose that makes sense. I mean personally I think one flag would be fine (just raise a warning if the file exists and no flag was given), but I guess is not as explicit.

jasonmotha:
Yeah. There's way too much log output to actually see warnings. But that's an entirely different problem with the architecture...

claudio525 (approved):

PR #353, opened 2021-02-11
 Add installation logic to workflow. Similar to perturbation install logic

jpa198:
Add required qpqs parameters to emod3d_defaults. These will be ignored if not used
Remove unused emod3d_defaults parameters

#### Checklist

### DIFF
--- a/scripts/cybershake/install_cybershake.py
+++ b/scripts/cybershake/install_cybershake.py
@@ -1,6 +1,6 @@
 import argparse
 from datetime import datetime
-import os
+from os import path
 
 from qcore.constants import (
     TIMESTAMP_FORMAT,
@@ -16,13 +16,11 @@ from shared_workflow.platform_config import platform_config
 AUTO_SUBMIT_LOG_FILE_NAME = "install_cybershake_log_{}.txt"
 
 
-def main():
-    logger = qclogging.get_logger()
-
+def load_args(logger):
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "path_cybershake",
-        type=str,
+        type=path.abspath,
         help="the path to the root of a specific version cybershake.",
     )
     parser.add_argument(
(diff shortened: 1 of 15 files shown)


jpa198:
Jenkins failing on changes to e3d.par and qcore dependency

scripts/cybershake/install_cybershake.py
```diff
@@ -74,81 +72,93 @@ def main():
...
         help="Don't use velocity model perturbations. If this is selected any perturbation files will be ignored.",
     )
-
+    parser.add_argument(
+        "--vm_qpqs_files",
+        action="store_true",
+        help="Use generated Qp/Qs files. If this is selected all events/faults must have Qp and Qs files.",
+    )
+    parser.add_argument(
+        "--ignore_vm_qpqs_files",
```
jasonmotha, line 81: You can make these two arguments exclusive.

You could even store it in one variable. A nullable bool. True - use it, False ignore it and None - don't expect it

jasonmotha: https://riptutorial.com/python/example/6133/setting-mutually-exclusive-arguments-with-argparse

claudio525 (approved):

jasonmotha (approved):

templates/gmsim/20.4.1.4/emod3d_defaults.yaml
```diff
@@ -57,7 +59,5 @@ ts_xy: 1
 ts_xz: 0
 ts_yz: 0
 tzero: 0.6
+use_qpqs: 0
 vmodel_swapb: 0
-xseis: 0
-yseis: 0
-zseis: 0
```
shot0829, line 63: why was these originally here and now we are removing them from all gmsim template?

shot0829 (approved):

PR #355, opened 2021-02-14
Update order of db states

jpa198:
#### Checklist


jasonmotha (commented):
Can we generate this automatically from the "constant"?

shot0829:
> Can we generate this automatically from the "constant"?

potentially we can, but that means the UI will be locked in with whatever enum that status was given.
Not sure if that is a good idea. 
Pros are that it removes the need to update query script manually if we have any update on qcore.constants
Cons are that it will have a lot less flexibility to the UI

shot0829 (approved):
As @jpa198 mentioned, this is 'patch/fix' PR, instead of a feature change PR.
I personally dont have a issue with this getting merged first, and we can update this later on.

claudio525 (approved):

PR #357, opened 2021-02-24
Adv im output test

shot0829:
#### Checklist


scripts/test_adv_IM.py
```diff
@@ -14,23 +24,81 @@
...
+    for model in adv_im_model:
+        # check if csv exist
+        csv_path = os.path.join(sim_dir, "IM_calc/{}.csv".format(model))
+        im_dir = get_im_calc_dir(sim_dir)
         # check file exist first to prevent crash
         try:
             df = pd.read_csv(csv_path)
         except FileNotFoundError:
             failed.append(model)
             print("csv for {} does not exist".format(model))
```
jpa198, line 53: `continue` so you don't need the following block to all be indented?

scripts/test_adv_IM.py
```diff
@@ -14,23 +24,81 @@
...
+
+            # if csv exist:
+            # check station count, if match, skip rest of checks
+            csv_station_count = len(df.station.unique())
+            # glob for station folders
+            list_folders = [x for x in os.listdir(im_dir) if os.path.isdir(x)]
+            if csv_station_count == len(list_folders):
+                # station folder count matches csv.station.count
+                continue
+            else:
```
jpa198, line 67: Either negate the test and remove the continue, or remove the else and de-indent the rest of the function

scripts/test_adv_IM.py
```diff
@@ -14,23 +24,81 @@
...
+                for station_dir in list_folders:
+                    # check if folder is a station run folder
+                    station_model_dir = os.path.join(station_dir, model)
+                    station_name = os.path.basename(station_dir)
+                    if not os.path.isdir(station_model_dir) or (
+                        station_name in df.station.unique()
+                    ):
+                        # does not match naming for a station run, possibly created by something else, skipping
+                        continue
+                    else:
```
jpa198, line 77: Again, negate and drop continue or drop the else and de-indent

scripts/test_adv_IM.py
```diff
@@ -14,23 +24,81 @@
...
+                        if result != True:
+                            # logs shows not all analysis failed to converge, but station failed to create csv
+                            print(f"{sim_dir}: {station_name} failed to run on {model}")
+                            failed.append(model)
     if len(failed) != 0:
-        print("some csv has empty data. models: {}".format(failed))
-        sys.exit(1)
+        print("some runs have failed,. models: {}".format(failed))
+        if exit:
+            sys.exit(1)
```
jpa198, line 95: Is there a case where we want a failing test to have a return code of 0?
Should just be able to make it always return 1
As there is no --exit argument, this would help any automated testing too

scripts/test_adv_IM.py
```diff
@@ -14,23 +24,81 @@
...
+                                check_status(component_outdir, check_fail=True)
+                                and result
+                            )
+                        # check of "Failed"
+                        # no "Failed" in all Analysis*.txt == crashed
+                        # "Failed" in all Analysis*.txt == failed to converge (normal), continue
+                        if result != True:
+                            # logs shows not all analysis failed to converge, but station failed to create csv
+                            print(f"{sim_dir}: {station_name} failed to run on {model}")
+                            failed.append(model)
```
jpa198, line 91: I suggest using a break in here, instead of continuing to check after we know it's failed e.g.
`if not check_status: print(); append(); break`

jasonmotha: What I'd recommend is breaking this into separate functions. It seems like this function has too much going on. What is the logical flow of the program and go from there. 

At least a comment in the docstring saying what testing steps are being done here.

jasonmotha:
This pull-request is also a great size. Easy to review.

scripts/test_adv_IM.py
shot0829: the reason of not using a break is that it shows you all the stations that had failed the analysis. So you can look at them all at once. otherwise it can potentially waste a lot of CH if you only realized that there is another station crashed after you run the whole event.

scripts/test_adv_IM.py
shot0829: the idea here is when you run the script directly, it behaves like a one-time scripts: show you the result and a related return code.
But if you are importing it, you can choose to not exit (otherwise it will also break the parent script that calls it)
So it will be handy if you have another wrapper that import this function and needs to run it multiple times.

scripts/test_adv_IM.py
shot0829: whats the issue of it being indented? I think logic is more clear if following try, except, else?

scripts/test_adv_IM.py
shot0829: wouldn't it be clearer to someone that reads the code to know what test these lines are doing what which test has higher priority by having the if, else and indent?
As these lines has no practical impact on the script nor compatibility issues, i dont see it as an issue? 
the computation time should be exactly the same as well.

scripts/test_adv_IM.py
jasonmotha: Is there currently any scripts importing it? If there isn't, I'm on the side of that this is introducing complexity.

scripts/test_adv_IM.py
jpa198: With all the extra (unneeded) indents, you go up to 8 indentation levels, which I think is excessive

jasonmotha: If you are unsure why these are unneeded - have a quick chat to James. He's got an idea of how you don't need as many indentation levels.

scripts/test_adv_IM.py
jpa198: If you drop the else and de-indent then the intent will be clear, because anyone reading it will know the effect continue has
Negating the test is probably less clear as it's harder to track down where the loop ends.
I'm still in favour of dropping the else and indentation level though

jasonmotha: In general computation time isn't a factor. It's more about being able to clearly see the progression / layout of the code.

scripts/test_adv_IM.py
jpa198: Ah, I forgot about the print statements (going to a slurm out file) having all the details from within a run

scripts/test_adv_IM.py
jpa198: Using break will not stop after the first station. It will stop after the first broken component for each station.
The bit is saves on is calling check_status after something has already failed.

I agree with breaking this function up though

shot0829: the line you commented on is looping at station level, instead of comp within a station; hence the confusion. 
but i do agree on breaking when one component fails.

jpa198 (dismissed):

scripts/test_adv_IM.py
jasonmotha: @jpa198 / @shot0829 Unresolving because I'd still like an answer to my question.

shot0829: @jasonmotha yes, the current script i used to wrap around it is bash, thats why the exit code is required for it to catch the result. 
but as i mentioned, if there's anyone decided to use the main function within their code, the unconditioned exit is going to cause trouble.
I personally think its a good idea too have a option to turn off potential problematic features.

shot0829: i think what i can do is to make it sys.exit by default, but return code when desired.

shot0829: such as 
sys.exit(1) if exit else return 1

scripts/verify_adv_IM.py
```diff
@@ -0,0 +1,113 @@
+# script to quickly check the status of adv_im for a event
+# checks for existance of aggregated csv
+# checks for nan/null in the csv
+# checks station in csv matches stations ran
+
+
+import argparse
```
claudio525, line 7: Nitpicking, but can thoughts on changing the order of the imports as per, https://www.python.org/dev/peps/pep-0008/#imports?
I'm def a fan of this one, and think it would be a good idea to this across the team.

Also use docstring notation for the script docstring above? (https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)

scripts/verify_adv_IM.py
```diff
@@ -0,0 +1,113 @@
...
+def main(sim_dir, adv_im_model, components, exit=True):
+
+    failed = []
+
+    for model in adv_im_model:
+        # check if csv exist
+        csv_path = os.path.join(sim_dir, "IM_calc/{}.csv".format(model))
+        im_dir = get_im_calc_dir(sim_dir)
+        # check file exist first to prevent crash
+        try:
```
claudio525, line 74: Any particular reason for using a try/except here instead of an if/else?

claudio525 (commented):
Some minor nitpicking

scripts/verify_adv_IM.py
shot0829: just to catch the error from pandas and continue if the file is missing.
without it, script will need to first use isfile to check the existence before reading it, otherwise a FileNotFoundError will exit the whole script.

scripts/verify_adv_IM.py
shot0829: I agree. i've re-ordered them and change the comments to docstring.

jasonmotha (approved):
This is definately one of my favourite PRs in terms of size, changes made and good feedback / discussion. Great work everyone.

claudio525 (approved):

claudio525 (approved):

jpa198 (approved):

PR #360, opened 2021-03-05
Move args default value to only be run by master

jpa198:
This prevents a crash when running HF

#### Checklist

### DIFF
--- a/scripts/hf_sim.py
+++ b/scripts/hf_sim.py
@@ -203,8 +203,8 @@ if __name__ == "__main__":
             # invalid arguments or -h
             comm.Abort()
 
-    if args.sim_bin is None:
-        args.sim_bin = binary_version.get_hf_binmod(args.version)
+        if args.sim_bin is None:
+            args.sim_bin = binary_version.get_hf_binmod(args.version)
 
     if is_master:
         logger.debug("=" * 50)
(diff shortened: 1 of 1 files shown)


jasonmotha (approved):

scripts/hf_sim.py
```diff
@@ -203,8 +203,8 @@ def args_parser(cmd=None):
             # invalid arguments or -h
             comm.Abort()
 
-    if args.sim_bin is None:
-        args.sim_bin = binary_version.get_hf_binmod(args.version)
+        if args.sim_bin is None:
+            args.sim_bin = binary_version.get_hf_binmod(args.version)
```
jasonmotha, line 207: Are you merging with the block above / below too?

scripts/hf_sim.py
```diff
@@ -203,8 +203,8 @@ def args_parser(cmd=None):
...
 
     if is_master:
```
claudio525, line 209: Can make this one `if is_master check`?

scripts/hf_sim.py
jpa198: Done

jasonmotha (approved):

shot0829 (approved):

claudio525 (approved):

jasonmotha:
Just curious. What actual bug does this fix? From what I can see in the code, because args gets broadcast a few lines down all processes should have the same copy of args?

jpa198:
> Just curious. What actual bug does this fix? From what I can see in the code, because args gets broadcast a few lines down all processes should have the same copy of args?

For all processes but rank 0, args = None, therefore args.sim_bin doesn't exist and throws an exception. 
This happens in MPI code, cue Bad Things.

PR #362, opened 2021-03-12
removed modification of vm_params in install_shared.py

shot0829:
#### Description
this part of changes was removed from the adv_IM PR. and is related  https://github.com/ucgmsim/Pre-processing/pull/173
#### Checklist


shared_workflow/install_shared.py
```diff
@@ -98,7 +98,9 @@ def install_simulation(
     }
 
     # VM params
-    vm_params_dict = utils.load_yaml(vm_params_path)
+    vm_params_dict = utils.load_yaml(
```
jasonmotha, line 101: Are we using the vm_params path variable anymore?

shot0829: line 112 
https://github.com/ucgmsim/slurm_gm_workflow/blob/392a9f80bd68c5852a89175d86739447b6539ffc/shared_workflow/install_shared.py#L112
still use it. 
the parent script that calls this function still pass this as a argument ( and is incorrect one in the test, using params_vel instead of the newest config format)
So there is two options:
1. remove the logic from parent script, so the config is always read from vel_mod_dir
    less flexible, but stable
2. leave the logic in parent script and remove it from this scripts. whoever calls install script should handle the path correctly.
    more flexible, but prone to mismatch crashes

shot0829: the vm_params_path in SimParams are not being used in install either, its serves as recording purpose.
so it should be re-directed to the value/choice whatever we decided to use from the two option i mentioned.

jpa198: We can just refactor vm_params_path out and replace it with sim_struct.get_vm_yaml(vm_dir).
"Mr Huang, tear down this parameter!"

jasonmotha: Yeah. VM params is defined by the simulation structure. If that moves, then we have other issues.

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -229,31 +223,25 @@ def install_fault(
...
 
+        if (
+            root_params_dict is None
+            or fault_params_dict is None
+            or sim_params_dict is None
+            or vm_params_dict is None
+        ):
+            # Something has gone wrong, returning without saving anything
+            logger.critical(f"Critical Error some params dictionary are None")
+            return
```
jasonmotha, line 234: I was about to say it might be nice to print something if it's an error like this. But tbh that was the original code behaviour and I've never seen this error before. So all good!

shared_workflow/install_shared.py
```diff
@@ -110,7 +96,7 @@ def install_simulation(
         SimParams.sim_dir.value: sim_dir,
         SimParams.srf_file.value: srf_file,
         SimParams.vm_params.value: vm_params_path,
-        SimParams.sim_duration.value: sim_duration,
+        SimParams.sim_duration.value: vm_params_dict[SimParams.sim_duration.value],
```
jasonmotha, line 99: Is sim duration stored in both vm_params and sim_params?

shot0829: currently yes. We store most of the params from vm_params into sim_params after install, instead of reading it from vm_params when needed. My change was just removing the hard-coded naming/string to use a centralized definition.

shot0829: by tracing the code a bit, it should be removed, as for some reason this was the only parameter that is still stored in sim_params. all other params are stored in vm_params ( and the way we read sim_params is set to read the vm_params by default as well)
At the time of the commit i was just trying to remove the hardcoded naming as i mentioned.

scripts/cybershake/install_cybershake_fault.py
jpa198: This is only going to raise if the vm_params nt doesn't match with the nt calculated from the sim duration and the gmsim version root_params dt value. I'm not sure this part of the PR needs to go in?
It doesn't seem to change the functionality (Other than the logging statement)

shot0829: this kind of error only happened in ancient times when we are not using proper config reading. what i did was simply moved it a few line above, and removed the lines that sets them all to None, if another check failed.
The feature @jpa198  referring to is line 236~243, now it simply returns instead of setting dictionary to None (this was usefully when we were still using imports instead of load_yaml)

shot0829: so these line now serve as checking if the dictionaries are None, instead of a code exit.

jpa198: If this only happened when we were importing dicts instead of loading yamls, then why is it still here?
I was actually referring to line 142(old)/112(new) of install_shared. 236 relates to comparing flo, which now has its own return

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -195,7 +194,7 @@ def install_fault(
             root_params_dict,
             fault_params_dict,
             sim_params_dict,
-            vm_add_params_dict,
+            vm_params_dict,
```
jpa198, line 197: Please remove the return, this is already loaded in install fault

scripts/cybershake/install_cybershake_fault.py
```diff
@@ -229,31 +223,25 @@ def install_fault(
             ignore_vm_qpqs_files=ignore_vm_qpqs_files,
         )
 
+        if (
+            root_params_dict is None
+            or fault_params_dict is None
+            or sim_params_dict is None
+            or vm_params_dict is None
```
jpa198, line 230: Don't need to check vm_params_dict as load_yaml shouldn't be able to return None

jpa198 (changes requested):

shared_workflow/install_shared.py
```diff
@@ -236,7 +205,7 @@ def install_simulation(
             else:
                 sim_params_dict.update({key: value})
 
-    return root_params_dict, fault_params_dict, sim_params_dict, vm_params_dict
+    return root_params_dict, fault_params_dict, sim_params_dict
```
jpa198, line 208: Line 142/111 (old/new) returns 4 Nones, there should only be 3, or it should just raise an exception

jasonmotha (approved):

jpa198 (approved):

PR #363, opened 2021-03-12
Clean up submit_IM_calc and add ims to the root defaults template

jpa198:
Breaks compatibility with old installs
Draft with the current direction I'm going with it
#### Checklist


scripts/submit_sim_imcalc.py
```diff
@@ -89,237 +36,146 @@ def submit_im_calc_slurm(
...
+            command_options[const.SlBodyOptConsts.simple_out.value] = "-s"
 
-    submit_yes = (
-        True if options_dict["auto"] else confirm("Also submit the job for you?")
-    )
+        if params["ims"][const.RootParams.extended_period.name]:
+            command_options[const.SlBodyOptConsts.extended.value] = "-e"
+            period_count = 100
+        else:
+            period_count = 15
```
jasonmotha, line 122: 31 / 131 But yeah not generally a fan of this. Unsure how to do it better though.

jasonmotha (commented):
*thumbsup*

scripts/submit_sim_imcalc.py
jpa198: Oh, pSA_periods is a variable I can take the length of. Getting the EE period count as a variable would probably require importing from IM_calc though

jasonmotha: That sounds promising. Don't you also need to import from IM_calc at some point anyway? Or is it separated well enough?

jpa198: Tada, just needed an import from qcore.constants

jasonmotha:
Wouldn't we be better off upgrading the Jenkins version of black?

jpa198:
> Wouldn't we be better off upgrading the Jenkins version of black?

Probably, we should get Sung to do it when there aren't many PRs going, then give each branch an 'update black' PR
It's a pip package, so there shouldn't be any platform issues

jpa198:
@jasonmotha  Nope, looks like it's fine, and my local black was just being weird. It's all fine now, and no upgrades needed anywhere

jpa198:
Turns out this is the whole PR, my original refactor did most of the leg work

jpa198:
Qcore PR must go in first

claudio525 (approved):

scripts/cybershake/auto_submit.py
```diff
@@ -56,6 +55,8 @@ def submit_task(
     if not os.path.isdir(ch_log_dir):
         os.mkdir(ch_log_dir)
 
+    params = utils.load_sim_params(os.path.join(sim_dir, "sim_params.yaml"))
```
shot0829, line 58: although its unlikely to change, but wouldn't it be better if we use the defined structure from qcore.simulation_structure?

jpa198: Good point, changed

scripts/obs_im_calc.sl
```diff
@@ -23,7 +23,7 @@ do
     if [[  `find ${D} -name accBB | wc -l` -ge 1 ]]
     then
         fault_name=`basename $D`
-        python $gmsim/workflow/scripts/im_calc_checkpoint.py ${obs_dirs}/IM_calc/ $((`ls $D/*/*/accBB | wc -l` / 3)) --event_name ${fault_name} --observed
+        python $gmsim/workflow/scripts/im_calc_checkpoint.py ${obs_dirs}/IM_calc/ $((`ls $D/*/*/accBB | wc -l` / 3)) 1 --event_name ${fault_name} --observed
```
shot0829, line 26: to make it easier for future usage without looking into the checkpoint script what '1' stands for,
i suggest making the comp a variable.
for example
```suggestion
       comp="000 090 geom"
       comp_count=`echo $comp | awk '{print NF}' `
       python $gmsim/workflow/scripts/im_calc_checkpoint.py ${obs_dirs}/IM_calc/ $((`ls $D/*/*/accBB | wc -l` / 3)) $comp_count --event_name ${fault_name} --observed
```
and 
```suggestion
        time python ${IMPATH}/calculate_ims.py $D/*/*/accBB a -o ${obs_dirs}/IM_calc/ -np ${SLURM_CPUS_PER_TASK} -i ${fault_name} -r ${fault_name} -c $comp -t o -e -s
```
default should still be geom BTW, just using 3 comps to show how it works.

shot0829 (approved):

claudio525 (approved):

PR #366, opened 2021-04-18
Adv im cleanup

shot0829:
#### Checklist


scripts/submit_sim_imcalc.py
```diff
@@ -109,9 +109,6 @@ def submit_im_calc_slurm(
             const.SlBodyOptConsts.advanced_IM.value
         ] = f"-a {body_options['models']} --OpenSees {qconfig['OpenSees']} "
 
-        print(
-            f"test {params[const.SlBodyOptConsts.advanced_IM.value]['match_obs_stations']}"
-        )
```
jpa198: logger.debug should replace the prints, rather than removing them. Perhaps a bit more context around the strings being logged too.
Same below

jpa198 (changes requested):

templates/adv_im_calc.sl.template
```diff
@@ -45,12 +43,14 @@ if [[ $success == 0 ]]; then
...
-    fd_name=`python -c "from qcore import utils; params = utils.load_sim_params('{{sim_dir}}/sim_params.yaml'); print(params.FD_STATLIST)"`
+    fd_name=$(getFromYaml "{{sim_dir}}/sim_params.yaml" FD_STATLIST)
     fd_count=`cat $fd_name | wc -l`
 
     # log metadata
-    python $gmsim/workflow/metadata/log_metadata.py {{sim_dir}} $step_name cores={{np}} fd_count=$fd_count models="{{models}}"start_time=$start_time end_time=$end_time status="COMPLETED"
+    python $gmsim/workflow/metadata/log_metadata.py {{sim_dir}} $step_name cores={{np}} fd_count=$fd_count models="{{models}}" start_time=$start_time end_time=$end_time status="COMPLETED"
+    # tar the temporary outputs from analysis
+    # following IM_calc_dir/station/model/component_dir
+    find {{sim_IM_calc_dir}}/*/*/ -mindepth 1 -maxdepth 1 -type d -exec bash -c 'tar --remove-files -uvf {{sim_IM_calc_dir}}/adv_im_out.tar "{}" ' \;
```
jasonmotha, line 53: I believe we didn't just want to tar them as it creates a lot of output not just n_files?

jasonmotha (changes requested):

scripts/submit_sim_imcalc.py
```diff
@@ -124,9 +127,7 @@ def submit_im_calc_slurm(
                     "got more than one folder globbed. please double check the path to the match_obs_stations is correct."
                 )
                 sys.exit()
```
jasonmotha, line 129: Will this exit close autosubmit too? If so, should print something to the console too.

jasonmotha (changes requested):

scripts/submit_sim_imcalc.py
jpa198: logger.error prints to the console, as well as to the log file

jasonmotha: Easy!

scripts/link_obs.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# link BB.bin from a previously completed runs
+# $1 src_root_dir: the root directory of a previous automated simulation runs
+# $2 des_root_dir: the root directory of a clean install of simultion runs
+# $3 list_fault: a list to events to link
+
+# IMPORTANT: structure is for validation runs only.
+# TODO: translate into python and incorporate qcore.simluated_structure for flexibility
```
jasonmotha, line 9: Does this do the same as the above python script?

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+                else fault
+            )
+            src_bb_path = simulation_structure.get_bb_bin_path(
+                simulation_structure.get_sim_dir(source_root_dir, realisation_name)
+            )
+            des_bb_path = simulation_structure.get_bb_bin_path(
+                simulation_structure.get_sim_dir(destination_root_dir, realisation_name)
+            )
+            makedirs(path.dirname(des_bb_path), exist_ok=True)
+            cmd = ["ln", "-s", f"{src_bb_path}", f"{des_bb_path}"]
```
jpa198: Use https://docs.python.org/3/library/os.html#os.symlink instead of calling run?

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+
+def main(source_root_dir, destination_root_dir, fault_selection_list):
+    faults = {}
+    with open(fault_selection_list, "r") as fault_file:
+        for line in fault_file.readlines():
+            fault, count, *_ = line.split()
+            count = int(count[:-1])
+            faults.update({fault: count})
+    for fault, count in faults.items():
+        for rel_no in range(count):
```
jasonmotha, line 36: range(start, end+1)

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    args = parser.parse_args()
+
+    return args
+
+
+def main(source_root_dir, destination_root_dir, fault_selection_list):
+    faults = {}
+    with open(fault_selection_list, "r") as fault_file:
```
jpa198: There's a function for this:
`from qcore.formats import load_fault_selection_file`

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+from qcore import simulation_structure
+from qcore import constants
+from shared_workflow.shared_automated_workflow import add_to_queue
+
+
+def arg_parser():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "source_root_dir", type=str, help="path to previous completed vaildation run"
```
jpa198: Setting these to type 'str' is redundant? abspath or Path would be better?

jasonmotha (changes requested):
I'd like a separate folder for the scripts below rather than putting them in the scripts folder. The scripts folder is way too hard to find what you are looking for.

Aside, this new functionality should be a separate pull-request rather than making this pull-request bigger and hence take longer in the review stage.

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
+import argparse
+from os import path, makedirs
+from subprocess import run
+
+from qcore import simulation_structure
+from qcore import constants
+from shared_workflow.shared_automated_workflow import add_to_queue
+
+
+def arg_parser():
```
jpa198: Function name is too close to the module name. I recommend load_args inline with other repo scripts

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+    parser.add_argument(
+        "fault_selection_list", type=str, help="a file that contains list of faults"
+    )
+
+    args = parser.parse_args()
+
+    return args
+
+
+def main(source_root_dir, destination_root_dir, fault_selection_list):
```
jpa198: Should have a more descriptive name. Even just link_bb()

scripts/link_bb.py
shot0829: > I'd like a separate folder for the scripts below rather than putting them in the scripts folder. The scripts folder is way too hard to find what you are looking for.
> 
> Aside, this new functionality should be a separate pull-request rather than making this pull-request bigger and hence take longer in the review stage.

good call. i'll move them to a new PR, as the functions are unrelated to this one.

scripts/link_bb.py
```diff
@@ -0,0 +1,63 @@
...
+            queue_folder = simulation_structure.get_mgmt_db_queue(destination_root_dir)
+            run_name = realisation_name
+            process_type = constants.ProcessType.BB.value
+            status = constants.Status.completed.value
+            job_id = None
+            add_to_queue(queue_folder, run_name, process_type, status, job_id)
+
+
+if __name__ == "__main__":
+    args = arg_parser()
```
jpa198: Would prefer to have this stuff be in a main function, prevents args being global

jpa198 (changes requested):

scripts/link_bb.py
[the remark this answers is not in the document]
shot0829: > I'd like a separate folder for the scripts below rather than putting them in the scripts folder. The scripts folder is way too hard to find what you are looking for.
> 
> Aside, this new functionality should be a separate pull-request rather than making this pull-request bigger and hence take longer in the review stage.

I'll move these two scripts to a new PR, as they are not related to this one.

scripts/link_obs.sh
shot0829: i'll mark these as resolved for this PR. as its causing clutter.

jpa198 (dismissed):
As long as this still passes E2E test I'm happy with it

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
jasonmotha, line 1: This file should be in the same dir as the standard obs_im calc.

But I'm not sure if this is the right folder for both as these are generally automated scripts. But pure naming wise this seems like the right place for them.

shot0829: i would suggest moving the standard obs_im calc script to here as well.
as they can only be submitted on nesi at the moment, until we give it some treatment to adjust headers accordingly.

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
...
+        # check for status
+        # skip if completed
+        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`
+        if [[ $? == 0 ]];then
+            continue
+        fi
+        time python $IMPATH/calculate_ims.py $path_eventBB a -o $path_event_out -np 40 -i $event -r $event -t  o -e -a $adv_IM_model --OpenSees_path $opensees_bin
+        # test for completion 
+        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`
+        exit
```
jpa198, line 65: That doesn't look right...

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
...
+        run_im=0
+        # check for status
+        # skip if completed
+        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`
+        if [[ $? == 0 ]];then
+            continue
+        fi
+        time python $IMPATH/calculate_ims.py $path_eventBB a -o $path_event_out -np 40 -i $event -r $event -t  o -e -a $adv_IM_model --OpenSees_path $opensees_bin
+        # test for completion 
+        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`
```
jpa198, line 64: res never gets used?

scripts/nesi_scripts/run_adv_im_obs_maui.sl
shot0829: oops....i forgot to remove it.

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
...
+        adv_IM_models=`python -c "from qcore.utils import load_yaml; params=load_yaml('$root_params'); print(' '.join(params['advanced_IM']['models']));"`
+    else
+    #failed to find a model from config/yaml
+        continue
+    fi
+    # run for all Models
+    for adv_IM_model in $adv_IM_models;
+    do
+        # re-initialize flag
+        run_im=0
```
jpa198, line 55: Unused variable?

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
...
+for event in `cat $list_event | awk '{print $1}'`;
+do 
+    echo $event
+    path_eventBB=$obs_dir/$event/*/*/accBB
+    path_IM_calc=$obs_dir/IM_calc
+    path_event_out=$path_IM_calc/$event
+    # tests before starting analysis if the csv is already there
+    # get station count
+    station_count=`ls $path_eventBB | cut -d. -f1 | sort -u | wc -l`
+    if [[ $station_count -lt 0 ]]; then
```
jpa198, line 39: Less than 0? Shouldn't trigger?

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -0,0 +1,74 @@
...
+if [[ $# -lt 2 ]];
+then
+    echo "please provide 1. path to obs_dir 2. list of event names"
+    exit 1
+fi
+
+obs_dir=$1
+list_event=$2
+# use default binary if not given
+opensees_bin=${3:-/nesi/project/nesi00213/opt/maui/tmp/OpenSees}
```
jpa198, line 28: /tmp/OpenSees? That path seems wrong

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -62,12 +62,13 @@ do
         time python $IMPATH/calculate_ims.py $path_eventBB a -o $path_event_out -np 40 -i $event -r $event -t  o -e -a $adv_IM_model --OpenSees_path $opensees_bin
         # test for completion 
         res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`
-        exit 
         if [[ $? == 0 ]];then
```
jasonmotha, line 65: Better too to compare res rather than $0

jpa198: Given that verify_adv_IM returns a string, and exits with a return code, this should be fine

shot0829: I prefer return code, unless we changed the output of the verify_adv_IM to be numerical, it creates extra dependencies.

jasonmotha: Ah yeah, I'm less keen to use $0 as it's led to bugs in the past in case lines are added between the call and the result.

jasonmotha: Unless, we get verify to print out the error and return a code?

Similar to how validate_vm is used?

jpa198: *$?

shot0829: adding a variable to store the return code solves that issue.

jpa198: Put the return code into its own variable, and use that in the check. Put them on adjacent lines and if anyone tries to put a line between them their PR gets denied

jpa198: Also a comment that the return code is needed for the test. So that someone else coming to this code knows why they shouldn't just jam a new line in there

jasonmotha: You could even put them on the same line with a ";"

jpa198: Too slow Jason, Jonney got there first

shot0829: was just talking to james saying that xD

jasonmotha: I said it 8 minutes ago...

scripts/nesi_scripts/run_adv_im_obs_maui.sl
shot0829: i've updated it to use qconfig

jasonmotha (dismissed):

jasonmotha:
E2E test / run of advim pending.

shot0829:
Doing another head to toe testing. (EMOD3D to BB and Adv_IM)

jasonmotha (approved):

scripts/submit_sim_imcalc.py
```diff
@@ -111,7 +111,7 @@ def submit_im_calc_slurm(
 
         # create temporary station list if "match_obs_stations" is directory
         if path.isdir(
-            params[const.SlBodyOptConsts.advanced_IM.value]["match_obs_stations"]
+            str(params[const.SlBodyOptConsts.advanced_IM.value]["match_obs_stations"])
```
jpa198, line 114: Why does this need to be cast to a str?

shot0829: isdir crashes if the variable is not a path/string format. casting it to str will allow the value to be None/null.

jpa198 (approved):

jpa198 (approved):

templates/adv_im_calc.sl.template
```diff
@@ -51,7 +51,7 @@ if [[ $success == 0 ]]; then
     # tar the temporary outputs from analysis
     # following IM_calc_dir/station/model/component_dir
     # using \; instead of + to avoid extra long exec bash command. (this will result in a exec for every file found
-    find {{sim_IM_calc_dir}}/*/*/ -mindepth 1 -maxdepth 1 -type d -exec tar --remove-files -uvf {{sim_IM_calc_dir}}/adv_im_out.tar -C {{sim_IM_calc_dir}} "{}" \;
+    find {{sim_IM_calc_dir}}/*/*/ -mindepth 1 -maxdepth 1 -type d -exec bash -c "f_path={}; tar --remove-files -uvf {{sim_IM_calc_dir}}/adv_im_out.tar -C {{sim_IM_calc_dir}} \${f_path#{{sim_IM_calc_dir}}/}" \;
```
jasonmotha, line 54: Is that #meant to be there?

That last one might be better as a separate variable - I can't work out what that guy is doing.

jpa198: Yeah, I think having that as a separate variable will help with clarity too, will help document what the variable is for

shot0829: the # is for sub-string removal.

i'll add in a new variable for clarity.

jasonmotha (approved):

jpa198 (approved):

PR #367, opened 2021-04-29
Metadata logging correctly. :face_with_head_bandage: 

jasonmotha:

### DIFF
--- a/templates/kisti_header.cfg
+++ b/templates/kisti_header.cfg
@@ -17,5 +17,5 @@ module add intel/18.0.3 impi/18.0.3 craype-network-opa craype-mic-knl
 ##export PYTHONPATH=$gmsim_root/Environments/nurion/workflow
 ##source $gmsim_root/Environments/nurion/virt_envs/python3_nurion/bin/activate
 export SLURM_JOB_ID="${PBS_JOBID/.pbs/}"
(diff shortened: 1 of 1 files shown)


templates/kisti_header.cfg
```diff
@@ -17,5 +17,5 @@ module add intel/18.0.3 impi/18.0.3 craype-network-opa craype-mic-knl
 ##export PYTHONPATH=$gmsim_root/Environments/nurion/workflow
 ##source $gmsim_root/Environments/nurion/virt_envs/python3_nurion/bin/activate
 export SLURM_JOB_ID="${PBS_JOBID/.pbs/}"
-export SLURM_NTASKS=$PBS_TASKNUM
+export SLURM_NTASKS=$(qstat $SLURM_JOB_ID -f | grep "used.ncpus" | cut -d"=" -f2 | xargs)
```
shot0829, line 20: why is it piped to xargs with no command?

jpa198 (dismissed):
Pending Jonneys question.
Otherwise, as long as it works empirically, I'm happy

templates/kisti_header.cfg
jasonmotha: To trim the space off the front. Apparently that's the best way to do that.

shot0829: for this case using 'bc' should do the trick as well, it will also raise a error if the format has unexpected change and are no longer numbers from the 'cut command'.

shot0829 (approved):

jpa198 (approved):

PR #368, opened 2021-05-03
fixed case where job not found on pbs queue

jasonmotha:


scripts/schedulers/pbs.py
```diff
@@ -17,15 +17,17 @@ def get_metadata(self, db_running_task: SchedulerTask, task_logger: Logger):
...
-        cmd = f"qstat -f -F json -x {db_running_task.job_id}"
-        out, err = self._run_command_and_wait(cmd, shell=True)
-        tasks_dict = json.loads(out)["Jobs"]
-        assert (
-            len(tasks_dict.keys()) == 1
-        ), f"Too many tasks returned by qstat: {tasks_dict.keys()}"
-        task_name = list(tasks_dict.keys()[0])
-        task_dict = tasks_dict[task_name]
         try:
+            cmd = f"qstat -f -F json -x {db_running_task.job_id}"
```
claudio525, line 21: I have some questions/suggestions for this code in both its original and current form, assuming I'm not just failing to understand it.

Firstly the try block, did nothing (imo) that should require exception handling (in the original version), it looked up some keys in a dictionary and did some replaceing?, neither of which should require a try/except block? So my guess is that the try/except was used to handle the case that keys were missing in the dictionary, which therefore meant:
`# a special case when a job is cancelled before getting logged in the scheduler`

Assuming my understanding isn't completely wrong here, I would suggest at the minimum to change the generic except block to change to be specific for missing key exception (KeyError?, ValueError? not sure tbh).

However, I would suggest to change this to a slightly more logical structure (in my opinion),

Have a try/except for the _run_command_and_wait as that seems to me like its the only block that should require one?
And then use an if statement (check for key existance in task dict) to check for the special case that was previously handled by the generic exception block?

jpa198 (approved):
Looks good. Any exceptions would mean the API has interface has changed and we need to look into it

claudio525 (approved):

jpa198 (approved):

PR #369, opened 2021-05-05
Add a script to test emod3d subdomain sizes for a given velocity mode…

jpa198:
…l and core count

Add a second script to process values from an e3d.par file
Added uses of this during submission to increase core count until there are no issues and in the emod3d template to crash early if there will be a problem from manual changes

#### Checklist


jasonmotha:
Haven't read through the math yet, general structure looks okay. Just mentioning that as I always say. The scripts folder is a mess so probably better off creating a subfolder. As your new file is importable I wouldn't be opposed to creating a new folder which is a module and starting to organise new stuff into there.

scripts/emod3d_scripts/__init__.py
```diff
@@ -0,0 +1,15 @@
+import argparse
```
jasonmotha, line 1: Would be nice to include as part of the docstring how this works.

jpa198: Ah, that was my pycharm template dropping itself in where it doesn't belong. Will remove.

jasonmotha (approved):

scripts/submit_emod3d.py
```diff
@@ -49,6 +49,8 @@ def main(
         nt = int(float(params.sim_duration) / float(params.dt))
 
         target_qconfig = get_machine_config(args.machine)
+        
+        retries = args.retries if hasattr(args, "retries") else None
```
jasonmotha, line 53: Why don't you just refactor out the args argument into this function? Seems simpler than the alternatives?

jasonmotha (commented):
wHEN YOU GET A SEC can you walk me through the math? Or have a reference wiki page?

scripts/submit_emod3d.py
jpa198: This is not a valid English sentence. Please refactor.

jasonmotha: [You have recieved a review request from @jasonmotha ]

jpa198:
> wHEN YOU GET A SEC can you walk me through the math? Or have a reference wiki page?

The main reason there isn't a wiki page or other documentation is that this is ripped straight out of EMOD3D. I (c/sh)ould probably do a wiki page validating my output against C code. But other than that it's a black box. I can do a quick walk through though.

scripts/emod3d_scripts/check_emod3d_subdomains.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Ported from the source of emod3d. Casting enforces C like behaviour.
+    Calculates the number of processes to use along each axis of a velocity model.
+    The argument min_nproc sets the unit size, allowing for blocks of grid points to be assigned to subdomains, instead of individual points
+    The nproc_x/z argument are available to mimic options available in the C. Not normally used
+    :param nproc: The number of processes to be used.
+    :param globnx: The number of velocity model grid points along the x axis.
+    :param globny: The number of velocity model grid points along the y axis.
+    :param globnz: The number of velocity model grid points along the z axis.
+    :param min_nproc: Multiplyer to perform calculations using cubes of min_nproc, defaults to 1.
```
jasonmotha, line 73: multiplier

scripts/emod3d_scripts/check_emod3d_subdomains.py
```diff
@@ -0,0 +1,210 @@
+"""
+Contains functions related to the calculation of emod3d subdomain boundaries.
+Functions ported from C contain a number of calls to np.int32 and np.float32 calls to emulate single precision integer and floating point behaviour.
+Code ported from emod3d v3.0.8 misc.c. This is consistent with v3.0.7. 
+While v3.0.4 uses long doubles in place of floats, this does not seem to practically increase the accuracy of calculation.
+"""
```
jasonmotha, line 6: Include stricter than necessary compliance but should be okay.

jasonmotha (changes requested):

jasonmotha (approved):

shot0829 (approved):

PR #370, opened 2021-05-17
Im entry

shot0829:
#### Checklist

### DIFF
--- a/scripts/cybershake/install_cybershake.py
+++ b/scripts/cybershake/install_cybershake.py
@@ -49,6 +49,7 @@ def load_args(logger):
     parser.add_argument(
         "--components",
         choices=list(Components.iterate_str_values()),
+        nargs="+",
         default=None,
         help="list of components to run IM_calcs, overwrites value in selected gmsim version template",
     )
@@ -176,7 +177,7 @@ def main():
             vm_qpqs_files=args.vm_qpqs_files,
             ignore_vm_qpqs_files=args.ignore_vm_qpqs_files,
             keep_dup_station=args.keep_dup_station,
-            components=components,
+            components=args.components,
             logger=qclogging.get_realisation_logger(logger, fault),
         )
 
--- a/scripts/cybershake/install_cybershake.py
+++ b/scripts/cybershake/install_cybershake.py
@@ -7,6 +7,7 @@ from qcore.constants import (
     ROOT_DEFAULTS_FILE_NAME,
     PLATFORM_CONFIG,
     HF_DEFAULT_SEED,
+    Components,
 )
 from qcore import qclogging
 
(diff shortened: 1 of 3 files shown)


shared_workflow/install_shared.py
```diff
@@ -77,6 +79,12 @@ def install_simulation(
     root_params_dict[RootParams.stat_vs_est.value] = vs30_file_path
     root_params_dict[RootParams.stat_vs_ref.value] = vs30ref_file_path
     root_params_dict["hf"][RootParams.seed.value] = seed
+    if components != None:
+        if not set(components).issubset(set(Components.iterate_str_values())):
+            message = f"{components} are not all in {Components}"
```
jasonmotha, line 84: This gets checked as part of IM_calc argparse. Is it worth having a double check here? Especially because parameters can change between install and run.

scripts/cybershake/install_cybershake.py
```diff
@@ -45,6 +46,13 @@ def load_args(logger):
         default="/nesi/project/nesi00213/StationInfo/non_uniform_whole_nz_with_real_stations-hh400_v20p3_land.ll",
         help="The path to the station info file path.",
     )
+    parser.add_argument(
+        "--components",
+        choices=list(Components.iterate_str_values()),
+        nargs="+",
+        default=None,
```
jasonmotha, line 53: Where's the default stored? Did we also need to update the 20.* templates?

jasonmotha (changes requested):

scripts/cybershake/install_cybershake.py
shot0829: james has already updated the templates with his previous PR.

shot0829: the question is do we want to have a deafult in the gmsim template and assume we will always have it there, or do we want a defautl value in the qcore libs.

shared_workflow/install_shared.py
shot0829: this check will raise an error at 'install' stage, instead of breaking when it reaches IM_calc step.
for a fully automated usage i think this is needed. and if the user decided to change parameters inbetween thats their own responsibility

sungeunbae (approved):
It will be nice to document that components=None (by default), but IM_Calculation will add "geom" component, if no component is specified..

jasonmotha:
> It will be nice to document that components=None (by default), but IM_Calculation will add "geom" component, if no component is specified..

The none in this means that it will go for the default in the root_params first of all.

sungeunbae:
> > It will be nice to document that components=None (by default), but IM_Calculation will add "geom" component, if no component is specified..
> 
> The none in this means that it will go for the default in the root_params first of all.

Cool - that part was not very clear. Makes sense.

shared_workflow/install_shared.py
```diff
@@ -77,6 +79,12 @@ def install_simulation(
     root_params_dict[RootParams.stat_vs_est.value] = vs30_file_path
     root_params_dict[RootParams.stat_vs_ref.value] = vs30ref_file_path
     root_params_dict["hf"][RootParams.seed.value] = seed
+    if components != None:
```
jpa198, line 82: Should use 'is not' when comparing against Bool/None values

jpa198 (changes requested):

jasonmotha (approved):

jpa198 (approved):

PR #374, opened 2021-06-10
added quotes on 000 to force it to be string

shot0829:
#### Checklist


jpa198 (approved):

jasonmotha:
What bug is this fixing?
Also, shouldn't it then be used for 090?

shot0829:
for some reason the yaml.load is treating 000 as int, while still treats 090 as string.

shot0829:
i can add quotes on the 090 as well, for safety measures.

jasonmotha (approved):
Oh yep. For the recent add component change.

jpa198 (approved):

PR #379, opened 2021-06-17
Update verify adv im

shot0829:
#### Checklist


scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,160 @@ def parse_args():
...
+    #
+    result_code = 0
+    for df, model_name in df_dict:
+        # check if any status >= 3 or != 0
+        if df["status"].ge(run_status.not_finished.value).any():
+            print(f"{model_name} have errors please check the status.csv")
+            result_code += 1
+        if df["status"].eq(run_status.not_started.value).any():
+            print(
+                f"{model_name} has some stations that havent been analysised. please check status.csv"
```
jpa198, line 207: analysed :P
Also capitals

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,160 @@ def parse_args():
...
+            )
+            result_code += 2
+        # sort index by status
+        df.sort_values("status", inplace=True, ascending=False)
+        # map status(int) to string before saving as csv
+        df["status"] = df["status"].map(lambda x: run_status(x).name)
+        status_csv_path = os.path.join(im_calc_dir, "{}_status.csv".format(model_name))
+        df.to_csv(status_csv_path, header=True, index=True)
+    return result_code
```
jpa198, line 216: Is this going to be useful once you've gone past one error? It'll give you some idea of the magnitude of the issues, but it'll still be a mix of +=1 and +=2 problems

scripts/verify_adv_IM.py
```diff
@@ -6,13 +6,47 @@
...
+class run_status(Enum):
+    not_started = 0
+    finished = 1
+    not_converged = 2
+    not_finished = 3
+    crashed = 4
+    unknown = 5
+
+
+column_names = [
```
jpa198, line 30: Globals should all caps names

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,163 @@ def parse_args():
...
+                station_component_status = run_status.finished.value
+            # check for 'Failed' keyword
+            elif check_status(component_outdir, check_fail=True):
+                # all logs showed "Failed", analysis was unable to converge
+                station_component_status = run_status.not_converged.value
+            elif time_list.count(None) == 0:
+                # if start and end are there, but no data = crashed
+                station_component_status = run_status.crashed.value
+            elif time_list.count(None) == 1:
+                # only starttime exist = wct timed out
```
jpa198, line 114: This makes the assumption that only the first value was found. The check should be made more explicit to verify this. e.g. time_list[0] is not None and time_list[1] is None.
With an error if the other case occurs

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,163 @@ def parse_args():
...
+                        {
+                            "station": station_list,
+                            "model": model,
+                            "component": component,
+                            "status": run_status.not_started.value,
+                        },
+                        columns=column_names,
+                    ),
+                ],
+                ignore_index=True,
```
jpa198, line 168: This can be refactored to use a list comprehension and one concat call. Off the top of my head concat is rather inefficient, so doing it once would be better

scripts/verify_adv_IM.py
shot0829: currently it doesnt have any usage. but it can be usefull if we have a external code that calls it and the error code can be easily identified.

shot0829: so by knowing the value, you can tell how many errors were there. ( i.e looking at it in binary, 00, 01, 10, 11)

jpa198: Yeah, but there's a for loop, so you might end up with 10 from 1+1 or from 2, or how ever many entries are in the df_dict

scripts/verify_adv_IM.py
```diff
@@ -6,13 +6,47 @@
```
jasonmotha, line 6: Would rename this script from verify to something else. Verify implies verification - rather than completeness checking. But I guess we do have hangovers like validateVM etc?

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,166 @@ def parse_args():
...
-                failed_stations.append(station_name)
-                # logs shows not all analysis failed to converge, but station failed to create csv
-                print(f"{station_dir} failed to run on {model}")
-                # break the loop as soon as one component fail the verification
-                break
-    return failed_stations
-
+            time_list = read_timelog(component_outdir)
+            if check_status(component_outdir):
+                # successed
```
jasonmotha, line 103: success (successed isn't a word lol)

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,166 @@ def parse_args():
...
+                station_component_status = run_status.finished.value
+            # check for 'Failed' keyword
```
jasonmotha, line 105: Be consistent with comments before or after conditional check (or even in-line)

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,166 @@ def parse_args():
...
+                )
+                for component in components
+            ],
+            ignore_index=True,
+        )
+
+        # a quick check to compare station count, will skip all other checks if successful.
+        if simple_check:
+            # using try/except to prevent crash
+            try:
```
jasonmotha, line 177: If we're only checking for filenotfound then we're probably better off using an if rather than a full on try/except? Thoughts?

scripts/verify_adv_IM.py
```diff
@@ -32,79 +66,166 @@ def parse_args():
...
+        print(model)
+        df_list.append((df_model, model))
+
+    #
+    result_code = 0b00
+    for df, model_name in df_list:
+        # check if any status >= 3 or != 0
+        if df["status"].ge(run_status.not_finished.value).any():
+            print(f"{model_name} have errors. Please check the status.csv")
+            result_code = result_code | 0b01
```
jasonmotha, line 207: Would be good to document the error codes somewhere.

jasonmotha (commented):
Looks good in general. Nice and simple.

scripts/verify_adv_IM.py
shot0829: using try/except to read the file prevents the very-unlikely-event that the file got removed (or IO failed) after the isfile check.

scripts/verify_adv_IM.py
shot0829: sorry i dont really get it. do you mean i should be consistent on if the comment for certain check is before or after the line?

jasonmotha: Yep. Especially because this one is duplicated below the line as well as above the line.

scripts/verify_adv_IM.py
```diff
@@ -6,13 +6,47 @@
...
 
 import pandas as pd
 
 from qcore import constants as const
-from IM_calculation.Advanced_IM.runlibs_2d import check_status
+from qcore.formats import load_station_file
+from IM_calculation.Advanced_IM.runlibs_2d import check_status, TIME_FORMAT, time_type
+
+
+class run_status(Enum):
```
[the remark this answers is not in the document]
shot0829: any suggestions? i named it that why because its a status of each opensees run. station_status?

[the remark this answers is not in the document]
jasonmotha: Whatever it's the status of? Like station_status etc.

Station status works as long as it's not like a per waveform thing.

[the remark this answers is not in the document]
jpa198: This is a component-station pair thing right? Maybe component_status would be sufficient?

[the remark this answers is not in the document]
shot0829: i've renamed it into analysis_status

scripts/verify_adv_IM.py
shot0829: the initial name was test_adv_im_status.py but it is triggering out automated testing ._.
so i had to find a word that is closer but not 'test'

jasonmotha: :laughing: That's perfect. I can't think of a better name / descriptor off the top of my head.

jpa198: check_adv_im.py? Otherwise it's fine for now, and we can change it later if there're issues

shot0829: i've moved the script to IM_calc repo, and renamed it check_adv_IM_status.py

jasonmotha: Perfect. That works for me!

templates/adv_im_calc.sl.template
```diff
@@ -31,7 +31,8 @@ start_time=`date +$runtime_fmt`
 end_time=`date +$runtime_fmt`
 
 # verify the result of the analysis
-res=`python $gmsim/workflow/scripts/verify_adv_IM.py {{sim_IM_calc_dir}} {{models}}`
+fd_name=$(getFromYaml "{{sim_dir}}/sim_params.yaml" FD_STATLIST)
+res=`python $gmsim/workflow/scripts/verify_adv_IM.py {{sim_IM_calc_dir}} {{models}} --station_file $fd_name`
```
jasonmotha, line 35: Update this path and we're good here.

jasonmotha (approved):

jpa198 (approved):

jpa198:
@shot0829 This isn't in yet?

PR #381, opened 2021-06-22
No args submit

sungeunbae:
#### Description
Removed args from main() function in 4 submit scripts.
#### Checklist
- Have you updated the README ? No 
- Have you updated the CHANGELOG? Yes


scripts/submit_emod3d.py
```diff
@@ -27,91 +27,103 @@
...
+    srf_name = Path(params.srf_file).stem
 
-        target_qconfig = get_machine_config(args.machine)
+    logger.debug("not set_params_only")
+    # get lf_sim_dir
+    sim_dir = params.sim_dir
+    lf_sim_dir = sim_struct.get_lf_dir(sim_dir)
 
-        retries = args.retries if hasattr(args, "retries") else None
+    # default_core will be changed is user passes ncores
```
jasonmotha, line 60: This comment doesn't seem in line with the code around it?

scripts/submit_emod3d.py
```diff
@@ -27,91 +27,103 @@
...
-                params.sim_dir,
-                srf_name,
-                target_machine=args.machine,
-                logger=logger,
-            )
 
 
 def get_lf_cores_and_wct(
-    logger, nt, params, sim_dir, srf_name, target_qconfig, ncore, retries=None
+    logger, nt, params, sim_dir, srf_name, target_qconfig, ncores, retries=None
```
jasonmotha, line 122: n_cores seems more appropriate? As it is technically two words right?

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
+    ncores: int = platform_config[const.PLATFORM_CONFIG.BB_DEFAULT_NCORES.name],
+    rel_dir: Path = Path("."),
+    retries: int = 0,
+    version: str = None,
+    write_directory: Path = None,
+    logger: Logger = get_basic_logger(),
+):
+    rel_dir = rel_dir.resolve()
+    try:
+        params = utils.load_sim_params(rel_dir / "sim_params.yaml")
```
shot0829, line 52: how does this string operation work? is it missing something? I never seen this before.

shot0829 (dismissed):

scripts/submit_bb.py
sungeunbae: https://realpython.com/python-pathlib/

shot0829: oh i see now, i did not realized that you changed the type to from basic os.path string to pathlib path.

jasonmotha (approved):
Few minor thoughts.

Happy if an E2E test passes.

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
 
 
-def main(args: argparse.Namespace, logger: Logger = get_basic_logger()):
-    params = utils.load_sim_params(os.path.join(args.rel_dir, "sim_params.yaml"))
-    ncores = platform_config[const.PLATFORM_CONFIG.BB_DEFAULT_NCORES.name]
+def main(
+    auto: bool = False,
+    machine: str = host,
+    ncores: int = platform_config[const.PLATFORM_CONFIG.BB_DEFAULT_NCORES.name],
+    rel_dir: Path = Path("."),
```
shot0829, line 44: so this is one thing i dont like about python, you cannot really force a argument to be a specific type and raise error instantly if incorrect type is parsed.

So here you are assuming the type to be pathlib.Path, but in the auto_submit.py it is using get_sim_dir, which returns a result of os.path.join ( string type )

Is there a PR for qcore that i'm not aware of? otherwise this seems like it will break, as a many of our code still uses simple string type.

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
+            bin = BBSeis(simulation_structure.get_bb_bin_path(params.sim_dir))
+        except:
+            logger.debug("Retried count > 0 but BB.bin is not readable")
+        else:
+            est_run_time_scaled = est_run_time * (retries + 1)
+
+    wct = set_wct(est_run_time_scaled, ncores, auto)
+    write_directory = write_directory if write_directory else params.sim_dir
+    if write_directory is None:
+        write_directory = params.sim_dir
```
shot0829, line 98: are these two line a duplicate of line 96?
since if write_directory is None should return False and use params.sim_dir

testing/test_submit_scripts/test_submit_emod3d.py
```diff
@@ -30,38 +28,34 @@ def test_main(set_up, mocker):
...
+        if str(x).find("emod3d_defaults.yaml") != -1
         else mocked_load_yaml(x),
     )
 
     for root_path, realisation in set_up:
-        args = get_input_params(
-            root_path, "{}_{}".format("submit_emod3d.py", function), params
+        rel_dir = (
+            Path(root_path)
+            / f"CSRoot/Runs/{get_fault_from_rel(realisation)}/{realisation}"
```
jpa198, line 46: Having a single string with '/' here defeats the purpose of using the Path join syntax.
This applies to test_submit_hf too

scripts/submit_bb.py
sungeunbae: Well spotted. Done

scripts/submit_emod3d.py
sungeunbae: Looks like someone didn't delete the comment 3 years ago.
```
 # default_core will be changed is user passes ncore
 n_cores = args.ncore
 if n_cores != default_core:
```
Removed the comment.

scripts/submit_bb.py
jpa198: In the current PR rel_dir is always passed in as a string. So either the type needs to be changed back to string, the variable cast to Path at the start of the function, or all references to main need to cast to Path before passing the value. This applies to hf, emod3d and post_emod3d too.
Converting sim_struct to use pathlib Paths is a good idea for a future refactor though, we'd have to either test the type or cast to Path at the start of every function though

sungeunbae: Good point. I looked at auto_submit.py and it relies too much on old os.path. 
Fixing all other references of os.path in auto_submit appears to be beyond the scope.
I'll allow rel_dir and write_directory to be passed as str.

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
+
+    underscored_srf = srf_name.replace("/", "__")
+
+    header_dict = {
+        "wallclock_limit": wct,
+        "job_name": f"bb.{underscored_srf}",
+        "job_description": "BB calculation",
+        "additional_lines": "###SBATCH -C avx",
```
jpa198, line 106: Given it's got three # it's not used by slurm, do we need this anymore? I'm not sure I've seen avx used since I started

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
+        body_template_params,
+        command_template_parameters,
+        add_args,
+    )
 
-        script_prefix = "{}_{}".format(sl_name_prefix, underscored_srf)
-        script_file_path = write_sl_script(
-            write_directory,
+    # Submit the script
+    submit_yes = True if auto else confirm("Also submit the job for you?")
```
jpa198, line 130: Do we need to use confirm here? We can use the argument to specify submit if we want to, otherwise just create the slurm/pbs file

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
         description="Create (and submit if specified) the slurm script for BB"
     )
 
-    parser.add_argument("--auto", nargs="?", type=str, const=True)
-    parser.add_argument(
-        "--version",
-        type=str,
-        default=platform_config[const.PLATFORM_CONFIG.BB_DEFAULT_VERSION.name],
-    )
+    parser.add_argument("--auto", action="store_true", default=False)
```
jpa198, line 156: This argument should be named --submit or something more descriptive. While it's being changed anyway...

jpa198 (changes requested):
A few things, most of them apply to all 4 submit scripts though I think

testing/test_submit_scripts/test_submit_emod3d.py
sungeunbae: Agreed. Going back to old faithful os.path and str

scripts/submit_bb.py
```diff
@@ -36,129 +37,178 @@ def gen_command_template(params):
...
+    logger: Logger = get_basic_logger(),
+):
+    rel_dir = Path(rel_dir).resolve()
+    try:
+        params = utils.load_sim_params(rel_dir / "sim_params.yaml")
+    except FileNotFoundError:
+        logger.error(f"Error: sim_params.yaml doesn't exist in {rel_dir}")
+        raise
+
+    params.sim_dir = Path(params.sim_dir).resolve()
```
jasonmotha, line 57: If we want to do this kind of transform I'd prefer it either stored as a local variable or as part of the load_sim_params. Modifying dictionaries (or namespaces) that are loaded from somewhere leads to potential confusion.

scripts/submit_emod3d.py
```diff
@@ -27,91 +27,103 @@
...
+    try:
+        params = utils.load_sim_params(rel_dir / "sim_params.yaml")
+    except FileNotFoundError:
+        logger.error(f"Error: sim_params.yaml doesn't exist in {rel_dir}")
+        raise
 
-    logger.debug("params.srf_file {}".format(params.srf_file))
-    # Get the srf(rup) name without extensions
-    srf_name = os.path.splitext(os.path.basename(params.srf_file))[0]
+    params.sim_dir = Path(params.sim_dir).resolve()
```
jasonmotha, line 47: Same comment as above. Especially when it is saved to a local variable a few lines down.

scripts/submit_bb.py
sungeunbae: Good point.

scripts/submit_bb.py
sungeunbae: This is historical when researchers preferred to interact with the sim configuration.  I'm removing it and renaming --auto  to --submit

scripts/submit_bb.py
sungeunbae: --submit argument is by default False, and it will take care of it.

scripts/submit_bb.py
sungeunbae: ok - agreed.

scripts/submit_emod3d.py
sungeunbae: done

scripts/submit_emod3d.py
```diff
@@ -44,9 +44,7 @@ def main(
         logger.error(f"Error: sim_params.yaml doesn't exist in {rel_dir}")
         raise
 
-    params.sim_dir = Path(params.sim_dir).resolve()
-
-    submit_yes = True if auto else confirm("Also submit the job for you?")
+    params_sim_dir = Path(params.sim_dir).resolve()
```
jasonmotha, line 47: would just `sim_dir` be more appropriate?

jasonmotha (changes requested):

jasonmotha (approved):
Code looks significantly more tidy! This is a really good change!

jpa198 (approved):

PR #383, opened 2021-06-29
new 100m config files (used on Kisti)

jasonmotha:


templates/gmsim/20.4.1.1/root_defaults.yaml
```diff
@@ -0,0 +1,20 @@
...
+bb:
+  fmin: 0.5
+  fmidbot: 1.0
+  no-lf-amp: True
+  dt: 0.005
+emod3d:
+   emod3d_version: 3.0.8
+flo: 1.0
+dt: 0.005
+pSA_periods: [0.01, 0.02, 0.03, 0.04, 0.05, 0.075, 0.1, 0.12, 0.15, 0.17, 0.2, 0.25, 0.3, 0.4, 0.5, 0.6, 0.7, 0.75, 0.8, 0.9, 1.0, 1.25, 1.5, 2.0, 2.5, 3.0, 4.0, 5.0, 6, 7.5, 10.0]
```
jpa198, line 18: Needs an im entry as per https://github.com/ucgmsim/slurm_gm_workflow/blob/master/templates/gmsim/20.4.1.2/root_defaults.yaml

jpa198 (changes requested):

templates/gmsim/20.4.1.1/root_defaults.yaml
jpa198: This line is now deprecated

jpa198 (approved):
Sure, looks ok
I do want to have an audit of the emod3d files at some point though. I know there a few that don't get used.

jasonmotha:
The audit is a good idea. You are probably the person with the best idea of EMOD3D params.

jpa198:
> The audit is a good idea. You are probably the person with the best idea of EMOD3D params.

Yeah, it should be a scoped ticket though, given how potentially breaking of a change it is. It would include set_runparams too, as that adds some junk too

shot0829 (approved):

PR #385, opened 2021-07-08
Enabled site specific 1D vm with site_vm_dir and site_specific arguments in fault_params.yaml

sungeunbae:
#### Description
Enabled site specific 1D vm with site_vm_dir and site_specific arguments in fault_params.yaml

#### Dependencies
Changes to qcore: Separate PR made
#### Checklist
- Have you updated the README : no
- Have you updated the CHANGELOG:  yes


jasonmotha (commented):
Few comments:
1)I'd prefer to keep it as the v1d dir rather than vm_dir as vm_dir generally means something something else.
2) Does the e2e tests still pass? I didn't have any issue with them on the old station list?
- the new one is incompatible with the old one.

sungeunbae:
> 2) Does the e2e tests still pass? I didn't have any issue with them on the old station list?
> 
> * the new one is incompatible with the old one.

e2e test can't pass unless qcore PR is merged in.

sungeunbae:
> 1)I'd prefer to keep it as the v1d dir rather than vm_dir as vm_dir generally means something something else.

All renamed.

jasonmotha:
e2e test should still be able to run under your environment with both branches checked out. e2e test is made to run in enviroments rather than master?

sungeunbae:
> e2e test should still be able to run under your environment with both branches checked out. e2e test is made to run in enviroments rather than master?

Ah, sorry. I was talking about jenkins test. e2e_test was broken already as it was unable to find stat_file. At least it runs now.  Updating e2e test to make it pass is beyond scope of this PR. Happy to revert the change.

jasonmotha:
> Ah, sorry. I was talking about jenkins test. e2e_test was broken already as it was unable to find stat_file. At least it runs now. Updating e2e test to make it pass is beyond scope of this PR. Happy to revert the change I made to .json files.

image
Yes please. The stat file exists for me. Ideally, we should be checking that the e2e test passes too but this often flies under the radar.

scripts/hf_sim.py
```diff
@@ -433,7 +446,7 @@ def run_hf(
             % (args.rvfac, args.rvfac_shal, args.rvfac_deep, args.czero, args.calpha),
             "%s %s" % (args.mom, args.rupv),
             args.stoch_file,
-            args.hf_vel_mod_1d,
+            velocity_model,
```
jasonmotha, line 449: I'd call this `v1d_path` for clarity.

scripts/hf_sim.py
```diff
@@ -301,8 +312,10 @@ def initialise(check_only=False):
                 dtype="f4",
             )
             # string parameters
-            if args.site_vm_dir != None:
-                vm = args.site_vm_dir
+            if args.site_v1d_dir != None:
+                vm = (
```
jasonmotha, line 316: Probably beyond scope but this variable `vm` refers to three different things depending on which function you are in. A bit of clarity around this would be awesome if possible.

scripts/hf_sim.py
sungeunbae: done

scripts/hf_sim.py
sungeunbae: done

jasonmotha:
For this one I'm waiting on qcore getting merged to see the test results before approving.

sungeunbae:
> For this one I'm waiting on qcore getting merged to see the test results before approving.

fair point

sungeunbae:
> For this one I'm waiting on qcore getting merged to see the test results before approving.

@jasonmotha  Happy?

jasonmotha (dismissed):

scripts/hf_sim.py
```diff
@@ -528,16 +543,18 @@ def validate_end(idx_n):
...
     in_stats = mkstemp()[1]
 
-    vm = args.hf_vel_mod_1d
+    v1d_path = args.hf_vel_mod_1d
     for s in range(work.size):
-        if args.site_vm_dir != None:
-            vm = os.path.join(args.site_vm_dir, "%s.1d" % (stations_todo[s]["name"]))
+        if args.site_v1d_dir != None:
+            v1d_path = os.path.join(
+                args.site_v1d_dir, "%s.1d" % (stations_todo[s]["name"].decode("ascii"))
```
jpa198, line 550: Should use an f string. Or at least .format

scripts/hf_sim.py
```diff
@@ -189,6 +197,11 @@ def args_parser(cmd=None):
 
     args = parser.parse_args(cmd)
 
+    if args.site_specific is True:
+        args.hf_vel_mod_1d = None
+    else:
+        args.site_v1d_dir = None
```
jpa198, line 203: As all the locations testing site_v1d_dir should be testing args.site_specific, we don't need to erase these arguments?
We should probably add some checking for site_v1d_dir without site_specific too, as that wouldn't do anything
Ideally we'd also only take one of hf_vel_mod_1d and site_v1d_dir, but I think that's a bit too much of a refactor for right now

jpa198 (changes requested):

scripts/hf_sim.py
```diff
@@ -528,16 +543,18 @@ def validate_end(idx_n):
     t0 = MPI.Wtime()
     in_stats = mkstemp()[1]
 
-    vm = args.hf_vel_mod_1d
+    v1d_path = args.hf_vel_mod_1d
     for s in range(work.size):
-        if args.site_vm_dir != None:
-            vm = os.path.join(args.site_vm_dir, "%s.1d" % (stations_todo[s]["name"]))
+        if args.site_v1d_dir != None:
```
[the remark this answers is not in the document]
sungeunbae: Replaced with "is not"

scripts/hf_sim.py
sungeunbae: Done

scripts/hf_sim.py
```diff
@@ -148,7 +148,15 @@ def args_parser(cmd=None):
...
     )
-    arg("-s", "--site-vm-dir", help="dir containing site specific velocity models (1D)")
+    arg(
+        "--site_specific",
+        action="store_true",
+        help="enable site-specific calculation",
+        default=False,
+    )
+    arg(
+        "-s", "--site_v1d_dir", help="dir containing site specific velocity models (1D)"
```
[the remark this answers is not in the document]
sungeunbae: Agreed

scripts/hf_sim.py
```diff
@@ -266,7 +279,7 @@ def initialise(check_only=False):
                     nl_skip,
                     ic_flag,
                     args.seed >= 0,
-                    args.site_vm_dir != None,
+                    args.site_v1d_dir != None,
```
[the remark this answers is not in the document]
sungeunbae: Sure.

scripts/hf_sim.py
```diff
@@ -301,12 +314,14 @@ def initialise(check_only=False):
                 dtype="f4",
             )
             # string parameters
-            if args.site_vm_dir != None:
-                vm = args.site_vm_dir
+            if args.site_v1d_dir != None:
```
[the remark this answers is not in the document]
sungeunbae: True

scripts/hf_sim.py
sungeunbae: Removed. Instead, added comment "ignored --site-specific is set" for hf_vel_mod_1d.

scripts/hf_sim.py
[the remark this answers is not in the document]
jasonmotha: :laughing: Great pun Sung.

shared_workflow/shared_template.py
```diff
@@ -83,8 +83,12 @@ def generate_command(
     command_parts.append(command_template.format(**template_parameters))
 
     for key in add_args:
+        if (
+            add_args[key] is False
+        ):  # Don't add store_true type arg at all if False.(This repo has no use of store_false)
+            continue
```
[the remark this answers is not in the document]
sungeunbae: Good idea. It sounds like we need to give default None to all store_true type arguments that will rely on write_sl_script()?

scripts/hf_sim.py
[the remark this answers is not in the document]
jpa198: :facepalm:

shared_workflow/shared_template.py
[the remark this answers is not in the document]
jpa198: Actually, I think this will be fine. If we do end up needing to pass False as a value in future we can look at changing our keywords then.
No changes needed here

jpa198 (approved):
Yeah, looks good

jpa198 (approved):

sungeunbae:
@jasonmotha  Will you have another look?

scripts/hf_sim.py
```diff
@@ -528,16 +540,18 @@ def validate_end(idx_n):
...
     in_stats = mkstemp()[1]
 
-    vm = args.hf_vel_mod_1d
+    v1d_path = args.hf_vel_mod_1d
     for s in range(work.size):
-        if args.site_vm_dir != None:
-            vm = os.path.join(args.site_vm_dir, "%s.1d" % (stations_todo[s]["name"]))
+        if args.site_v1d_dir is not None:
+            v1d_path = os.path.join(
+                args.site_v1d_dir, f"{stations_todo[s]['name'].decode('ascii')}.1d"
```
jasonmotha, line 547: I'd prefer if this was called with `str(stations_todo[s]['name'])` Thoughts @jpa198 

I've just been bitten by bytestrings too much before to be comfortable with changes to/from decode.

scripts/hf_sim.py
```diff
@@ -548,7 +562,11 @@ def validate_end(idx_n):
         validate_end(work_idx[-1] + 1)
 
     os.remove(in_stats)
-    print("Process %03d of %03d finished (%.2fs)." % (rank, size, MPI.Wtime() - t0))
+    print(
+        "Process {} of {} completed {} stations ({:.2f}).".format(
```
jasonmotha, line 566: If you are changing it, why not a f-string? Just curious?

shared_workflow/shared_template.py
```diff
@@ -83,8 +83,12 @@ def generate_command(
...
         command_parts.append("--" + key)
-        if add_args[key] is True:
+        if add_args[key] is True:  # store_true type arg needs no value
```
jasonmotha, line 91: Do we need the `is True`.
Sorry if this has been asked / clarified already?

scripts/hf_sim.py
sungeunbae: Just copied from the log below

shared_workflow/shared_template.py
sungeunbae: without "is True" add_args[key] having any non-negligible value will be regarded as True

jasonmotha (approved):

PR #386, opened 2021-07-14
Adv im fix

shot0829:
#### Description
1.updated the station matching logic so that is match both ways, this address the bug that when observed data have stations that was not included in simulation (thus failing the check_status).

2. The result of check_status is also set to flush to job_id.out so its easier to check the result, instead of logging in mgmt_db, which require more steps to check the output. the 
DB now logs the error code, instead of full message.

#### Checklist

### DIFF
--- a/scripts/nesi_scripts/run_adv_im_obs_maui.sl
+++ b/scripts/nesi_scripts/run_adv_im_obs_maui.sl
@@ -46,7 +46,7 @@ do
     do
         # check for status
         # skip if completed
-        res=`python $IMPATH/../Advanced_IM/check_adv_IM_status.py $path_event_out $adv_IM_model`; res_return_code=$?
+        python $IMPATH/../Advanced_IM/check_adv_IM_status.py $path_event_out $adv_IM_model; res_return_code=$?
 
         # return code from verify_adv_IM is used to determine status.
         if [[ $res_return_code == 0 ]];then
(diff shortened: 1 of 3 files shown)


jasonmotha (commented):
Code looks good. Can we have a quick summary of what bug this was fixing / feature requested?

shot0829:
updated the Description to contain a bit more detail about the bug fixes

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -52,15 +46,15 @@ do
     do
         # check for status
         # skip if completed
-        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`; res_return_code=$?
+        python $IMPATH/../Advanced_IM/check_adv_IM_status.py $path_event_out $adv_IM_model; res_return_code=$?
```
jpa198, line 49: IMPATH should be changed to be the second level IM_calculation directory so that you don't have to back track here and just below?
The PYTHONPATH export can be adapted to work with the change too

shot0829: thats doable and easy to fix, the reason i did not changed that was to minimize the changes, as its not part of the issue.

jasonmotha: I'd change it over. Clean code is better than legacy code.

jasonmotha (changes requested):
As above.

sungeunbae (dismissed):

scripts/nesi_scripts/run_adv_im_obs_maui.sl
```diff
@@ -52,15 +55,15 @@ do
     do
         # check for status
         # skip if completed
-        res=`python $gmsim/workflow/scripts/verify_adv_IM.py $path_event_out $adv_IM_model`; res_return_code=$?
+        python $IMPATH/../Advanced_IM/check_adv_IM_status.py $path_event_out $adv_IM_model --stations ${stations[@]}; res_return_code=$?
```
jpa198, line 58: This is still having to backtrack?

shot0829: updated the recent commit

jasonmotha (approved):

jpa198 (approved):

PR #387, opened 2021-07-16
Prepro auto jason

jasonmotha:
#### Description

Adding VM generation to the automated workflow as an optional entry point. Currently functionality is maintained.

I haven't had a thorough review of my code so there is still some tidying to be done. I've already put a comment in on something I want fixed.

As this is a high-impact change, I've got three reviewers.


shared_workflow/install_shared.py
```diff
@@ -108,15 +108,15 @@ def install_simulation(
...
-    nt = float(vm_params_dict["sim_duration"]) / root_params_dict["dt"]
-    if not isclose(nt, round(nt)):
-        logger.critical(
-            "Simulation dt does not match sim duration. This will result in errors during BB. Simulation duration must "
-            "be a multiple of dt. Ignoring fault. Simulation_duration: {}. dt: {}.".format(
-                vm_params_dict["sim_duration"], root_params_dict["dt"]
-            )
-        )
-        return None, None, None
+    # nt = float(vm_params_dict["sim_duration"]) / root_params_dict["dt"]
```
jasonmotha, line 111: UNCOMMENT THESE LINES JUST PUT SOME SAFETY AROUND IT. to: future me from: a very kind soul

templates/task_config_prepro.yaml
```diff
@@ -0,0 +1,6 @@
+VM_PARAMS: ONCE
+VM_GEN: ONCE
+# if you select the VM_PERT option you need to re-install (install_cybershake.py) after generating perturbations to
+# enable the perturbation to take effect
+VM_PERT: NONE # ALL
+INSTALL_FAULT: ONCE
```
jpa198, line 6: You didn't add the "no_vm_pert" step? I would have expected that to be easy enough to add, and give install the dual dependency

jpa198 (approved):
A few areas for improvement, but looks ok. Still needs that commented bit sorted out too though

sungeunbae (approved):
Looks good as long as @jpa198's comments are all addressed.

scripts/nesi_scripts/install_fault.sl
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+# script version: slurm
+#
+# must be run with sbatch install_fault.sl [realisationDirectory] [OutputDirectory] [managementDBLocation]
```
shot0829, line 4: seems to not match the args from below?

scripts/cybershake/auto_submit.py
```diff
@@ -307,6 +313,96 @@ def submit_script_to_scheduler(script_name, target_machine=None, **kwargs):
...
+                    {
+                        "VM_PARAMS_YAML": str(
+                            Path(sim_struct.get_fault_VM_dir(root_folder, run_name))
+                            / "vm_params.yaml"
+                        ),
+                        "STAT_FILE": str(params.stat_file),
+                        "FAULT_DIR": fault_dir,
+                        "FDSTATLIST": str(Path(fault_dir) / f"fd{str(params.sufx)}.ll"),
+                        "MGMT_DB_LOC": root_folder,
+                        "REL_NAME": run_name,
```
shot0829, line 400: since the same key-string is repeating multiple times, and for the unlikely senario that the value/string of the keys are changed, would it be a good idea to have a mapping somewhere?

shot0829: after reading a bit more i realized that these keys really dont serve any purpose currently. the keys are not used in the scheduler at all.

scripts/nesi_scripts/install_fault.sl
```diff
@@ -0,0 +1,65 @@
...
+
+#SBATCH --job-name=install_fault
+#SBATCH --time=00:15:00
+#SBATCH --cpus-per-task=1
+
+if [[ -n ${CUR_ENV} && ${CUR_HPC} != "mahuika" ]]; then
+    source $CUR_ENV/workflow/install_workflow/helper_functions/activate_env.sh $CUR_ENV "mahuika"
+fi
+VM_PARAMS=$1
```
shot0829, line 13: i would suggest adding a quick check to make sure all arguments are not empty.
such as 
${1:?please provide the path to VM_PARAMS } 
for all of them, just in case there was a unexpected change in the script above.

scripts/nesi_scripts/vm_gen.sl
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# script version: slurm
+#
+# must be run with sbatch vm_gen.sl [realisationDirectory] [OutputDirectory] [managementDBLocation]
```
shot0829, line 4: same issue as the other *.sl script

scripts/nesi_scripts/vm_gen.sl
```diff
@@ -0,0 +1,68 @@
...
+
+#SBATCH --job-name=VM_GEN
+#SBATCH --time=01:00:00
+#SBATCH --cpus-per-task=32
+
+if [[ -n ${CUR_ENV} && ${CUR_HPC} != "mahuika" ]]; then
+    source $CUR_ENV/workflow/install_workflow/helper_functions/activate_env.sh $CUR_ENV "mahuika"
+fi
+
+VM_PARAMS_YAML=$1
```
shot0829, line 14: same suggestion as previous script

scripts/nesi_scripts/vm_params_gen.sl
```diff
@@ -0,0 +1,67 @@
...
+
+#SBATCH --job-name=VM_PARAMS
+#SBATCH --time=00:15:00
+#SBATCH --cpus-per-task=1
+
+if [[ -n ${CUR_ENV} && ${CUR_HPC} != "mahuika" ]]; then
+    source $CUR_ENV/workflow/install_workflow/helper_functions/activate_env.sh $CUR_ENV "mahuika"
+fi
+
+REL_CSV=$1
```
shot0829, line 14: with this number of arguments required, i would not only suggest having a empty-value check, but also put down a 

if [[ ! $# != 9 ]]

to make sure the numbers are correct.

scripts/nesi_scripts/vm_pert.sl
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+# script version: slurm
+#
+# must be run with sbatch vm_gen.sl [realisationDirectory] [OutputDirectory] [managementDBLocation]
```
shot0829, line 4: same not matching issue.

scripts/nesi_scripts/vm_pert.sl
```diff
@@ -0,0 +1,70 @@
...
+#SBATCH --cpus-per-task=4
+
+if [[ -n ${CUR_ENV} && ${CUR_HPC} != "mahuika" ]]; then
+    source $CUR_ENV/workflow/install_workflow/helper_functions/activate_env.sh $CUR_ENV "mahuika"
+fi
+
+module load FFTW
+module load GCC/7.4.0
+
+VM_PARAMS_YAML=$1
```
shot0829, line 17: as suggest before.

shot0829 (approved):

jpa198 (approved):
Noice

scripts/cybershake/auto_submit.py
```diff
@@ -113,7 +116,7 @@ def submit_script_to_scheduler(script_name, target_machine=None, **kwargs):
             {
                 "XYTS_PATH": os.path.join(
                     sim_struct.get_lf_outbin_dir(sim_dir),
-                    "{}_xyts.e3d".format(run_name.split("_")[0]),
+                    "{}_xyts.e3d".format(run_name),
```
sungeunbae, line 119: why is split no longer needed?

scripts/cybershake/install_cybershake.py
```diff
@@ -70,6 +70,12 @@ def load_args(logger):
         action="store_true",
         help="Keep stations if they snap to the same grid-point",
     )
+    parser.add_argument(
+        "--no_check_vm",
+        action="store_false",
+        dest="check_vm",
+        help="Set this flag if you are generating VMs from the automated workflow",
+    )
```
sungeunbae, line 78: I personally find no_XXX switch with store_false option confusing. Can we make it positive, like "skip_vm_check" and store_true? You could be also a bit more explicit by setting the default value.

testing/test_shared_workflow/test_install_shared.py
```diff
@@ -16,6 +16,7 @@ def test_install_simulation(set_up):
                 ("CSRoot", "AdditionalData", "PangopangoF29/")
             ):
                 input_params[i] = os.path.join(root_path, input_params[i])
+        input_params.insert(10, False)
```
sungeunbae, line 19: Comment please

sungeunbae (commented):
I think I can approve it as it is, but would like to get the answer to my questions

scripts/cybershake/auto_submit.py
jasonmotha: https://github.com/ucgmsim/slurm_gm_workflow/commit/41afab43be6a16d7c5e956092a4a4e64ae284282#diff-23c9da3698676160c79b0fe584cfc923adcc553e412af402d8f1078f9516ff7bL195

Run name (aka e3d filenames) became the realisation name rather than the fault name.

scripts/cybershake/install_cybershake.py
jasonmotha: The aim is to maintain existing functionality, hence why adding an argument to skip the vm checking was added.

The reason for the store_false is to avoid checks later that state `if not skip_vm_check` as that at a glance is a lot harder to read than `if check_vm`.

I'll add a default value for clarity.

sungeunbae: Sure - I was thinking we could still set check_vm = not skip_vm_check, and keep the rest of the code unchanged, but I can accept it as it is - nice to have the default value.

jpa198 (approved):

sungeunbae (approved):

PR #391, opened 2021-08-04
Fix hf_sim bug

jpa198:
The previous string format style was "%s" % var, causing var to be converted to type string and then replacing the %s in the string.
The new type using "{:s}".format(var) instead expects var to be of type string, and raises an exception when this is attempted.
Instead leaving the brackets empty allows var to be cast to string and replace the {}.
Minimum example for the issue:
"{:s}".format(0.5)
Working equivalent:
"{}".format(0.5)

#### Checklist

### DIFF
--- a/scripts/hf_sim.py
+++ b/scripts/hf_sim.py
@@ -434,26 +434,26 @@ if __name__ == "__main__":
             str(args.sdrop),
             local_statfile,
             args.out_file,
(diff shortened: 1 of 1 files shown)


jpa198:
To be clear, most of the variables used in the part of the code with changes seem to be floats

shot0829 (approved):

shot0829 (approved):

jasonmotha (approved):

PR #392, opened 2021-08-04
Kisti update

shot0829:
#### Description
changed how module load/dels are handled.
Backward compatible, as its non-crashing if load fails.

https://wiki.canterbury.ac.nz/display/QuakeCore/Loading+Modules+for+specific+tools_dir+with+different+compilers

In pair with qcore PR:
https://github.com/ucgmsim/qcore/pull/242
#### Checklist


jasonmotha:
I'm not convinced this is the best way to handle modules. Why are we changing it from loading with the environment as we had done?

shot0829:
this should remove the need of complex rules in bash scripts, such as load_default_modules.sh and python_load_functions.sh. 
https://github.com/ucgmsim/slurm_gm_workflow/blob/0ceff85ecc829ed942b01223793a55a7c95b134b/install_workflow/load_default_modules.sh
These scripts are often not updated, and only works on NeSI.
With the proposed way of loading, we only need to keep qcore.config updated, and it will load the modules that are specifically written when binaries are compiled.

shot0829:
And as we run activate_env, it will load whatever modules needed for binaries in tools_dir, so the behavior on NeSI is almost the same, but instead of hard-coded in the bash script, it will now load it from the give path (module_requirement from qcore.config)

shot0829:
This also provide us the ability to flexibly change the tools_dir that was compiled using different compiler.
For example, on Nurion I had pre-compiled several with different combinations, and if one fails, we can quickly retreat to another (happened twice for the past year)
And this can be done easily by simply changing the symbolic link, or changing the path in qcore/configs/machines
CrayCC-8.6.3_impi
current -> Intel-19.0.5_impi
GCC-8.3.0_impi
GCC-8.3.0_openmpi
Intel-19.0.5_impi

shot0829:
> I'm not convinced this is the best way to handle modules. Why are we changing it from loading with the environment as we had done?

After re-read your commend, i realized that the change to NeSI and Tacc header was not necessary, as SLURM can inherit login-nodes environment without issue.
I've reverted the change to NeSI and Tacc header. But the change to Nurion (PBS) still applies

jasonmotha:
Sweet. Lots of good comments. I'm pretty convinced; but will chat more tomorrow about.

For reference, given the scope of the change; this should have been planned out better.

shot0829:
A side note: the reason this wasn't planned before head is that this was more or less a side-product of me trying to fix the Nurion environment.
As Nurion's MPI and modules wasnt as stable as NeSI's I had to try out most of the combination manaully, and i need something that doesnt require me to edit ten lines of code every time i switch tools_dir.

jpa198 (approved):

sungeunbae (approved):

PR #394, opened 2021-08-06
Install qcore like the old docker did

jpa198:
If we don't need qcore to be installed each time, then we can remove the clone line. 
It should be based on latest qcore though.

This is to counter a change made in #388 line 36 of the old jenkins file

#### Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
         mkdir -p /tmp/${env.JOB_NAME}/sample0
         cd /tmp/${env.JOB_NAME}
         git clone https://github.com/ucgmsim/qcore.git
-	PYTHONPATH=/tmp/${env.JOB_NAME}/qcore:$PYTHONPATH
+	PYTHONPATH=/tmp/${env.JOB_NAME}/qcore:${PYTHONPATH}
 
         ln -s $HOME/data/testing/slurm_gm_workflow/SGMW /tmp/${env.JOB_NAME}/build
         cd /tmp/${env.JOB_NAME}/sample0
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,8 +16,7 @@ pipeline {
         mkdir -p /tmp/${env.JOB_NAME}/sample0
         cd /tmp/${env.JOB_NAME}
         git clone https://github.com/ucgmsim/qcore.git
-	cd /home/jenkins/qcore
-	python setup.py install --user --no-data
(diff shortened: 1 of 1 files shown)


sungeunbae (approved):
If I remember correctly, installing qcore was under /tmp/slurm_gm_worklow/qcore then linked to somewhere less obscure when mounted onto docker's file system. If it is directly installed in /home/jenkins/qcore, I am afraid it could interfere with other testing that depends on qcore if executed  at the same time.

sungeunbae (changes requested):
I clicked a wrong button. Github on phone is fiddly!

jpa198:
> If I remember correctly, installing qcore was under /tmp/slurm_gm_worklow/qcore then linked to somewhere less obscure when mounted onto docker's file system. If it is directly installed in /home/jenkins/qcore, I am afraid it could interfere with other testing that depends on qcore if executed at the same time.

That's a valid issue. So in theory we should make a virtual environment for each repo, so that we can do this sort of installation without side effects

jasonmotha:
Who's going to be working on isolating the test environments? I guess that was a great thing about docker.

shot0829:
it will require going through all jenkin's file and add in creating virt_env line.
and wipe the virt_env if test passed.
might need a cron-job to clean-up the failed ones in a set period.

jpa198:
> it will require going through all jenkin's file and add in creating virt_env line.
> and wipe the virt_env if test passed.
> might need a cron-job to clean-up the failed ones in a set period.

There's a post step, we should always be able to run clean up right? We just never cleaned up the qcore download, so they always stuck around between jobs

sungeunbae:
> it will require going through all jenkin's file and add in creating virt_env line.
> and wipe the virt_env if test passed.
> might need a cron-job to clean-up the failed ones in a set period.

dont think it needs a virtual env. Just install under /tmp/slurm_gm_workflow/somehash/qcore and add it to PYTHONPATH and remove whole directory after the test. can be easily done altering Jenkinsfile

shot0829:
> > it will require going through all jenkin's file and add in creating virt_env line.
> > and wipe the virt_env if test passed.
> > might need a cron-job to clean-up the failed ones in a set period.
> 
> dont think it needs a virtual env. Just install under /tmp/slurm_gm_workflow/somehash/qcore and add it to PYTHONPATH and remove whole directory after the test. can be easily done altering Jenkinsfile

I agree, that is a lot more simpler, easy and fast to achieve.

jpa198:
Turns out it was already using the local qcore download, everything is fine, no need for this PR

PR #399, opened 2021-09-06
Added virtual env to Jenkins

sungeunbae:
#### Description
Tested OK - storage footprint seems ok
#### Checklist
- Have you updated the CHANGELOG  yes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,7 +61,7 @@ pipeline {
         always {
                 echo 'Tear down the environments'
                 sh """
-                    rm -rf /tmp/${env.JOB_NAME}/*
+#                    rm -rf /tmp/${env.JOB_NAME}/*
                 """
             }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
         always {
                 echo 'Tear down the environments'
                 sh """
-#                    rm -rf /tmp/${env.JOB_NAME}/*
+                    rm -rf /tmp/${env.JOB_NAME}/*
                 """
             }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,6 @@ pipeline {
                     rm -rf build
                     ln -s $HOME/data/testing/slurm_gm_workflow/SGMW build
                     echo "[ Linking test data ]"
-                    cd ${env.JOB_NAME}
                     rm -rf sample0
                     mkdir sample0
                     cd sample0
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,9 @@
 pipeline {
     agent any
+    
+    environment {
+        VIRT_ENV="/tmp/${env.JOB_NAME}/${env.ghprbActualCommit}/venv"
+    }
     stages {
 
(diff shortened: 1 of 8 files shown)


jpa198 (dismissed):

Jenkinsfile
```diff
@@ -1,48 +1,67 @@
...
+            steps {
+                echo "[[ Start virtual environment ]]"
+                sh """
+                    echo "[ Current directory ] : " `pwd`
+                    echo "[ Environment Variables ] "
+                    env
+# Each stage needs custom setting done again. By default /bin/python is used.
+                    source /var/lib/jenkins/py3env/bin/activate
+                    mkdir -p /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}
+# I don't know how to create a variable within Jenkinsfile (please let me know)
```
shot0829, line 15: are these comments suppose to be here? 😂 
if so, try these two method:
1. define it as a globle before the pipeline
using `def virtenv = /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}/venv`
2. inside script section:
`script {
    virtenv = /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}/venv
}`
using #2 requires  env.virtenv  if its going to be used in different steps.

Jenkinsfile
```diff
@@ -1,48 +1,67 @@
...
+                    python -m venv /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}/venv
+# activate new virtual env
+                    source /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}/venv/bin/activate
+                    echo "[ Python used ] : " `which python`
+                    cd ${env.WORKSPACE}
+                    echo "[ Install dependencies ]"
+                    pip install -r requirements.txt
+                    echo "[ Install qcore ]"
+                    cd /tmp/${env.JOB_NAME}/${env.ghprbActualCommit}
+                    rm -rf qcore
```
shot0829, line 26: if we are creating a new virtenv everytime, is this rm still needed?

shot0829, line 26: are these lines only used to have a up-to-date qcore?
just wondering why using rm and clone, instead of just git pull.
because after the most recent changes that added data to the repo, a full clone is going to take a lot longer than just a pull

sungeunbae: rm is done "just-in-case". In most cases, if it is cloned at /tmp/qcore/{env.ghprbActualCommit}, qcore should be empty. Sometimes, env.ghprbActualCommit can be a branch name (especially when manually triggered via Jenkins web interface), and in such a case qcore may not be empty. It's safer to delete whatever is there and pull the fresh clone.

Jenkinsfile
sungeunbae: Thanks - I'll give a go

sungeunbae: Ended up with environment directive - It only can keep one variable despite the documentation. Jenkins is such a joy. 
http://13.238.107.244:8080/job/slurm_gm_workflow/345/console

shot0829 (approved):

jpa198 (approved):

jasonmotha (approved):

PR #401, opened 2021-09-09
Pbs argparse

shot0829:
#### Checklist

### DIFF
--- a/scripts/schedulers/pbs.py
+++ b/scripts/schedulers/pbs.py
@@ -169,7 +169,7 @@ class Pbs(AbstractScheduler):
     @staticmethod
     def process_arguments(script_path: str, arguments: Dict[str, str]):
         """
-            keys in arguments must match whatever the pbs script is expecting, otherwise will fail
+        keys in arguments must match whatever the pbs script is expecting, otherwise will fail
         """
         # construct a string
         args_string = ""
--- a/scripts/schedulers/pbs.py
+++ b/scripts/schedulers/pbs.py
@@ -19,8 +19,11 @@ class Pbs(AbstractScheduler):
         """
         cmd = f"qstat -f -F json -x {db_running_task.job_id}"
 
-        out, err = self._run_command_and_wait(cmd, shell=True)
-        json_dict = json.loads(out)
+        out, err = self._run_command_and_wait(cmd)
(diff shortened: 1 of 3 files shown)


scripts/schedulers/pbs.py
```diff
@@ -20,7 +20,10 @@ def get_metadata(self, db_running_task: SchedulerTask, task_logger: Logger):
         cmd = f"qstat -f -F json -x {db_running_task.job_id}"
 
         out, err = self._run_command_and_wait(cmd, shell=True)
-        json_dict = json.loads(out)
+        # special treatment after nurion added GPU replated variable in qstat that is a single \ which will cause json parser unable to read
```
jasonmotha, line 23: replated?

Removes '\\' as nurion added a GPU ? that is a single '\' which causes json parser issues. 


In general comments should be active voice with the action at the start - I still can't figure out what replated is lol.

jasonmotha (commented):
.

scripts/schedulers/pbs.py
shot0829: typo..... it was supposed to be "related" 😂

jasonmotha (commented):
It feels like a few of the commands are a bit aggressively in the list rather than broken up. Do you need all the arguments to be in one argument?

shot0829:
> It feels like a few of the commands are a bit aggressively in the list rather than broken up. Do you need all the arguments to be in one argument?

it was mostly for consistancy.

jasonmotha:
> > It feels like a few of the commands are a bit aggressively in the list rather than broken up. Do you need all the arguments to be in one argument?
> 
> it was mostly for consistancy.

What do you mean by that?

shot0829:
> > > It feels like a few of the commands are a bit aggressively in the list rather than broken up. Do you need all the arguments to be in one argument?
> > 
> > 
> > it was mostly for consistancy.
> 
> What do you mean by that?

so all PBS calls have the same behavior. this will reduce the chance of any pbs parsing behavior catch us off guard.
I do realize that having everything a shell call have potential risks, but with how strict pbs is on the syntax, i dont think this will be a issue.

jasonmotha (approved):
Doesn't quite seem like best practice. But PBS is full of that.

sungeunbae (approved):

PR #403, opened 2021-09-16
I love when code gets tested when it changes argparse, ooh passive ag…

jasonmotha:
…gressive commit messages are also good

1/3 of Viktors bugs are in this PR, the other two are beyond my quickfix scope. I'll discuss the other two at standup.


jpa198 (approved):

sungeunbae (approved):
I love a concise commit message :)

PR #406, opened 2021-10-04
im_plot.sl: Find all components from IM.csv and plot them

sungeunbae:
#### Description

Currently im_plot.sl opens a CSV file, and assumes it is "geom" component for plotting. The assumption is mostly correct as "geom" is the default component. If user wanted to include rotd50 (or any components) in the IM csv, this fix will pick up the additional component and plot it.

#### Checklist
- Have you updated the README : No
- Have you updated the CHANGELOG : Not yet

### DIFF
--- a/scripts/im_plot.sl
+++ b/scripts/im_plot.sl
@@ -13,8 +13,7 @@ fi
 
 CSV_PATH=$1
 STATION_FILE_PATH=$2
-OUTPUT_XYZ_DIR=$3
-
+# $3 is assigned to OUTPUT_XYZ_DIR below
 SRF_PATH=$4
 MODEL_PARAMS=$5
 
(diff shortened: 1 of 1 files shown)


scripts/im_plot.sl
```diff
@@ -31,51 +30,61 @@ echo ___im plot___
...
-        fi
-        exit_val2=$?
-        if [[ $exit_val2 != 0 ]]; then
-            failed=1
-            failed_msgs+=( "failed $res2" )
-        else
-            success_msgs+=( "successfully $res2" )
-        fi 
-    done
+COMPS = `cat $CSV_PATH |cut  -d , -f 2 |sort |uniq` # find what components are used
```
jasonmotha, line 33: Originally, as per the slack clarification it was just adding RotD (if available) as it is going to be our component of interest going forward.

If we want to do multiple components, it might be worth specifying it in the config file - but beyond the scope of this PR.

In general, we wouldn't need the map plots for 000, 090, ver so seems overkill to generate these all the time if they are in the IM_csv

scripts/im_plot.sl
```diff
@@ -31,51 +30,61 @@ echo ___im plot___
...
+COMPS=($(echo $COMPS |tr " " "\n")) # convert to an array of strings
+COMPS_LEN=${#COMPS[@]} # length of the array
+for (( i=1;i<$COMPS_LEN;i++ )); #${COMPS[0]} is the header "component"
+do 
+    COMP=${COMPS[$i]}
+    OUTPUT_XYZ_DIR=$3/$COMP
```
jasonmotha, line 39: I'd prefer to have all argparse at the start of the script - just so it's in one place?

scripts/im_plot.sl
```diff
@@ -31,51 +30,61 @@ echo ___im plot___
...
-            failed=1
-            failed_msgs+=( "failed $res2" )
-        else
-            success_msgs+=( "successfully $res2" )
-        fi 
-    done
+COMPS = `cat $CSV_PATH |cut  -d , -f 2 |sort |uniq` # find what components are used
+COMPS=($(echo $COMPS |tr " " "\n")) # convert to an array of strings
+COMPS_LEN=${#COMPS[@]} # length of the array
+for (( i=1;i<$COMPS_LEN;i++ )); #${COMPS[0]} is the header "component"
```
jasonmotha, line 36: The skipping the first element doesn't seem robust to skip the header "component" `0` normally comes before `c`

Either way this comment is mitigated by the one two lines up.

scripts/im_plot.sl
```diff
@@ -31,51 +30,61 @@ echo ___im plot___
...
-       python $gmsim/workflow/scripts/cybershake/add_to_mgmt_queue.py $MGMT_DB_LOC/mgmt_db_queue $SRF_NAME IM_plot failed $SLURM_JOB_ID --error "$failed_msgs"
+        python $gmsim/workflow/scripts/cybershake/add_to_mgmt_queue.py $MGMT_DB_LOC/mgmt_db_queue $SRF_NAME IM_plot failed $SLURM_JOB_ID --error "$res"
     fi
-else
-    python $gmsim/workflow/scripts/cybershake/add_to_mgmt_queue.py $MGMT_DB_LOC/mgmt_db_queue $SRF_NAME IM_plot failed $SLURM_JOB_ID --error "$res"
-fi
+
+done
+
+
```
jasonmotha, line 87: Any reason for the extra whitespace?

jasonmotha (changes requested):
Big one being the discussion about all components vs specified vs geom + rotd50

scripts/im_plot.sl
sungeunbae: Is there a possibility when you (or who else?) do IM calc with 000 090, ver, geom and rotd50 all, then only wish to plot geom and rotd50?

scripts/im_plot.sl
sungeunbae: Good point - fixed in the new commit

scripts/im_plot.sl
jasonmotha: Conversation is slightly out of date but currently the default components calculated are:
image

Which is a bit more than what most runs need.

scripts/im_plot.sl
```diff
@@ -30,14 +32,18 @@ echo ___im plot___
...
+else 
+    comps_to_plot=(geom)
+fi
+
+for comp in ${comps_to_plot[@]};
 do 
-    COMP=${COMPS[$i]}
-    OUTPUT_XYZ_DIR=$3/$COMP
-    res=`python $gmsim/visualization/im/spatialise_im.py $CSV_PATH $STATION_FILE_PATH --out_dir $OUTPUT_XYZ_DIR -c $COMP`
+    OUTPUT_XYZ_DIR=OUTPUT_XYZ_PARENT_DIR/$comp
```
jasonmotha, line 45: `OUTPUT_XYZ_DIR=$OUTPUT_XYZ_PARENT_DIR/$comp` ? Are you missing a `$`

Just for sanity have you tested this with the automated workflow?

sungeunbae: Good catch. No I haven't tested it. I was going to ask you to test assuming you already have a setup.

jasonmotha: I don't have something that would easily be useable. I only really have the CS dir.

sungeunbae: ok - i'll take care of it

scripts/im_plot.sl
sungeunbae: Roger that

jasonmotha (approved):

jpa198 (approved):

PR #421, opened 2021-11-30
update estimate_wct to handle rotd in im_calc

shot0829:
#### Description
https://wiki.canterbury.ac.nz/display/QuakeCore/IM_calc+WCT+Estimation+update
#### Checklist

### DIFF
--- a/shared_workflow/platform_configs/platform_local.json
+++ b/shared_workflow/platform_configs/platform_local.json
@@ -12,6 +12,7 @@
   "PLATFORM_ACCOUNTS": ["nesi00213", "comm00213"],
   "DEFAULT_MEMORY" : "16G",
   "SCHEDULER": "bash",
+  "VM_PERT_DEFAULT_NCORES":1,
   "HEADER_FILE": "nesi_header.cfg",
   "AVAILABLE_MACHINES":  ["local"],
   "VELOCITY_MODEL_DIR": "/nesi/project/nesi00213/VelocityModel",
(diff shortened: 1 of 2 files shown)


jasonmotha (dismissed):

sungeunbae (approved):

jasonmotha (dismissed):

jpa198 (approved):

shared_workflow/platform_configs/platform_local.json
```diff
@@ -12,6 +12,7 @@
   "PLATFORM_ACCOUNTS": ["nesi00213", "comm00213"],
   "DEFAULT_MEMORY" : "16G",
   "SCHEDULER": "bash",
+  "VM_PERT_DEFAULT_NCORES":1,
```
jasonmotha, line 15: It's default set to 4 in the slurm script. Should we keep that?

shot0829: this is for local runs, which i doubt anyone would use?

shot0829: and all other values are set to 1 as well.

shot0829: i personally think its fine to keep it as one, as if anyone is using this config, that means
1. the workflow failed to identify the machine, thus safer to assume its serial
2. advanced user know what he is doing, thus should be able to adjust the config to required.

jasonmotha (approved):

PR #430, opened 2022-03-11
changes to use new nurion account x2319a02

sungeunbae:
#### Description
Changes needed to untie KISTI account hpc11a02 from the workflow

#### Dependencies
x2319a02 KISTI account

#### Checklist
- Have you updated the README (yes/no)? No
- Have you updated the CHANGELOG (yes/no)?  No


jpa198 (dismissed):

workflow/automation/org/kisti/vm_gen.pbs
```diff
@@ -37,7 +37,7 @@ echo python $gmsim/Pre-processing/VM/vm_params2vm.py -t 18 $REL_NAME $VM_PARAMS_
 python $gmsim/Pre-processing/VM/vm_params2vm.py -t 18 $REL_NAME $VM_PARAMS_YAML -o $OUTPUT_DIR
 
 chmod g+rwXs -R $OUTPUT_DIR
-chgrp nesi00213 -R $OUTPUT_DIR
+chgrp `groups` -R $OUTPUT_DIR
```
jasonmotha, line 40: `groups` prints out all groups that a user is part of. `id -gn $username` only prints the primary group. (unsure if this will be a personal group or the group group though)

workflow/calculation/gmsim_templates/20.4.1.2/root_defaults.yaml
```diff
@@ -1,5 +1,5 @@
 hf:
-  version: 5.4.5.3_new_gnu
```
jasonmotha, line 2: A sim version shouldn't change HF version (for backwards compatibility). If you need a different HF version, you should create a new gm_sim version.

jasonmotha (changes requested):
Couple of minor changes. But nice and simple.

workflow/automation/org/kisti/vm_gen.pbs
sungeunbae: This is a Nurion-specific change and a user is not a member of multiple groups.

workflow/calculation/gmsim_templates/20.4.1.2/root_defaults.yaml
sungeunbae: Fair point. Will revert this change. Out of curiosity, do we have 5.4.5.3_new_gnu on NeSI?

workflow/automation/org/kisti/vm_gen.pbs
jpa198: Experimentally they give the same result, however if one is guaranteed to return only one value, then we should use that one

sungeunbae: Sure. Fixed in the new commit, even though I know KISTI policy mandates only one group membership.

workflow/calculation/gmsim_templates/20.4.1.2/root_defaults.yaml
jasonmotha: image
Yep we do.

jasonmotha (approved):

jpa198 (approved):

PR #434, opened 2022-04-13
Site specific bb

jpa198:
#### Checklist


jasonmotha (approved):

jasonmotha (changes requested):
Thought this was a different PR...

sungeunbae:
Love to see a bit more description in this PR

workflow/calculation/bb_sim.py
```diff
@@ -380,67 +390,96 @@ def unfinished():
...
-            hf_c = np.hstack((hf_start_padding_ts, hf_acc[:, c], hf_end_padding_ts))
-            lf_c = np.hstack((lf_start_padding_ts, lf_acc[:, c], lf_end_padding_ts))
-            if is_master and i == 0:
-                if len(hf_c) != len(lf_c):
-                    logger.critical(
-                        "padded hf and lf have different number of timesteps, aborting. "
-                    )
-                    comm.Abort()
-            bb_acc[:, c] = (hf_c + lf_c) / 981.0
+            pga = np.max(np.abs(hf_acc), axis=0) / 981.0
```
sungeunbae, line 423: Nothing wrong here, but I'd like a comment mentioning LF/HF acc is in cm/s/s, unlike BB (which is in g). My heart was racing when I saw this line...and I calmed down when I double checked qcore.timeseries code.

sungeunbae (dismissed):

workflow/calculation/bb_sim.py
jpa198: Yeah, I've been having a few issues with that. LF/HF are cm/s/s, OpenSees needs m/s and BB is in g. Will make a comment on it though

jpa198:
I've integrated Chris and Felipes site specific amplification work here.
The site response script has an main() entry point for observed ground motions, while the simulated workflow imports it into BB as  an alternative to the vs30 based amplification.
The original site specific workflow has to remove the vs30 amplification first, this bypasses the need for deamplification.
The two test sites are included as examples for now. A more permanent location will need to be determined for large scale runs.

jpa198:
Things that still need to be determined:

- Storage location for site data
- Where the setting will be given to run site specific
    - Currently it will always be run for all sites with a site file
    - The easy option to make it optional is to put the site directory in gmsim versions with bb: site_specific: True/False
    - The default file folder is currently set in the organisation config. This should probably also have an option to over-ride it with a custom one, probably during install

jasonmotha (dismissed):

workflow/calculation/site_specific_BB/site_response.py
```diff
@@ -0,0 +1,442 @@
...
+
+    # integrate acceleration to get velocity (in m/s) for input to OpenSees
+    cm_to_m_multiplier = 1.0 / 100
+    bbgm_decon_vel = (
+        cumulative_trapezoid(bbgm_decon, dx=dt, initial=0) * cm_to_m_multiplier
+    )
+
+    with TemporaryDirectory() as td:
+        td = Path(td)
+        with NamedTemporaryFile(dir=td) as file_name:
```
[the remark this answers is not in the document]
jpa198: As it's going into a temporary directory it doesn't really matter if it's got a constant name or not, it'll be destroyed after running either way

workflow/calculation/bb_sim.py
```diff
@@ -389,51 +392,57 @@ def unfinished():
         )
         lf_acc = np.copy(lf.acc(stat.name, dt=bb_dt))
         hf_acc = np.copy(hf.acc(stat.name, dt=bb_dt))
-        station_yaml = os.path.join(args.site_specific_dir, f"{stat.name}.yaml")
-        if os.path.isfile(station_yaml):
+        station_yaml = os.path.join(str(args.site_specific_dir), f"{stat.name}.yaml")
```
jasonmotha: Hmm, is this worthwhile having in the if statement with a walrus?

Well, someway of not creating it if site_specific_dir isn't True

workflow/calculation/site_specific_BB/site_response.py
```diff
@@ -0,0 +1,443 @@
...
+    logger.info(f'got stdout line from subprocess: {subp.stdout.decode("utf-8")}')
+    logger.info(f'got stderr line from subprocess: {subp.stderr.decode("utf-8")}')
+    # TODO: Add debugging if it didn't work
+
+    out_file_path = out_dir / "out.txt"
+    return out_file_path
+
+
+def load_args():
+    parser = argparse.ArgumentParser(allow_abbrev=False)
```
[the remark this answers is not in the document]
jpa198: It's just a default I have on my python script template. Tries to reduce unintended consequences from incorrect arguments. Removed

jpa198:
Won't pass testing until https://github.com/ucgmsim/qcore/pull/264 is in

workflow/calculation/site_response_BB/site_template.yaml
```diff
@@ -0,0 +1,40 @@
+H_soil: 0.0  # Depth to bottom of soil column
+dampS_base: 0.0  # Vs damping ratio of layer to which you deconvolve
+
+numLayers: 0  # number of soil layers (not counting layer 0 which is bedrock)
+waterTable: 0.0  # if water not present set waterTable anywhere below depth of model
```
joelridden, line 5: Probably not a big thing but there is two types of styling here, camel case and some form of snake case

jpa198: These are hang overs from the source code I ported, editing these would require further TCL edits, which I'm trying to avoid (Never mind the refactor i did to get the variables out)

jpa198:
My test run has completed successfully, this can go in

jpa198:
:tada:

jasonmotha (approved):

joelridden (approved):

PR #440, opened 2022-05-11
removed vs_ref stat_file and unused parameter and the duplicate stat_…

jasonmotha:
…file entry

#### Checklist

### DIFF
--- a/workflow/automation/tests/test_shared_workflow/test_install_shared.py
+++ b/workflow/automation/tests/test_shared_workflow/test_install_shared.py
@@ -23,7 +23,7 @@ def test_install_simulation(set_up):
 
         # The pickle file was missing the 10th argument. This enables the vm to be checked.
         input_params.insert(10, True)
-        input_params.pop(8)  # Removes the vs_ref stat file parameter
+        input_params.pop(9)  # Removes the vs_ref stat file parameter
 
         test_output = install_shared.install_simulation(*input_params)
         root_params_dict = test_output[0]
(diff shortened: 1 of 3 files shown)


jasonmotha:
I found recently that stat_file was stored in both root_params and sim_params.

This PR removes the one in sim_params as it is a copy of the one in root params. And tidied up a couple of things.

workflow/automation/tests/test_shared_workflow/test_install_shared.py
```diff
@@ -23,6 +23,7 @@ def test_install_simulation(set_up):
 
         # The pickle file was missing the 10th argument. This enables the vm to be checked.
         input_params.insert(10, True)
+        input_params.pop(9)  # Removes the vs_ref stat file parameter
```
jpa198, line 26: Lol

jpa198 (approved):

workflow/automation/tests/test_shared_workflow/test_install_shared.py
sungeunbae: Can we fix the pickle file? It's getting multiple band-aids.

sungeunbae (changes requested):

workflow/automation/tests/test_shared_workflow/test_install_shared.py
jpa198: The test data is quite a mission to change at the moment. The current process involves logging into the test machine (I don't think I even have a working account on it), finding the right data, loading the data from pickle, editing it, saving back into pickle and running a test to make sure it works. Which is a lot of effort compared with adding a line or two to patch the test.
In the short term a few more careful patches won't hurt, however in the longer term I think testing needs a relatively large overhaul.

jasonmotha: @sungeunbae Thoughts?

sungeunbae (approved):

PR #452, opened 2022-08-30
Gmsimtemplates

jasonmotha:


sungeunbae:
We once discussed that a change to gmsim templates is not safe, as it makes the simulation that uses the particular version no longer reproducible.

sungeunbae:
I'm ok with a newly created version, but at least some explanation should be added to this PR to justify the change to existing versions.

jpa198:
The change to the existing templates seems to make them line up with validation defaults rather than cybershake defaults. Should we make 20.4.1.1_cs for the cybershake variants, so we don't have to manually specify them each time? It somewhat reduces the effectiveness of having template files if we have to override them each time we do a (non validation) run

jasonmotha:
If CS versions become more ubiquitious, let's create versions for them. But in the interim - I'm happy with manually changing things as needed. The main reason for the change is that more people are using the validation settings than the CS settings.

```[removed binary that no longer exists](https://github.com/ucgmsim/slurm_gm_workflow/pull/452/commits/bfee196eedf3414b36751fa280d1cef342f6050e)```
```[updated gmsim templates to the validation sesttings for VM generation](https://github.com/ucgmsim/slurm_gm_workflow/pull/452/commits/08fafa75ea848a82e0c5fff188779c6c2279dc13)```

Comments on changes as commit messages. =)

**Review by jpa198 — APPROVED** [review]

File: workflow/calculation/gmsim_templates/20.4.1.2/root_defaults.yaml [hunk 1]
```diff
@@ -1,5 +1,5 @@
 hf:
-  version: 5.4.5.3_new_gnu
+  version: 5.4.5.3
```
sungeunbae, line 2: @jasonmotha 
I remember you didn't like this change previously.
https://github.com/ucgmsim/slurm_gm_workflow/pull/430#discussion_r826479424

I also checked, hf binary that you state "no longer existis" is still present.

jpa198: Good catch, that should be consistent at least

jasonmotha: image
That HF binary is no longer on Kisti.

My comment about that earlier was more related to changing 5.4.5.3 to 5.4.5.4. If I recall correctly Viktor made those the new_gnu version functionally the same as the standard Rob Graves version.

jasonmotha: Sorry poor screenshot 
image

sungeunbae: @jasonmotha  As you pointed out earlier (see the screenshot), NeSI still has that binary, so backward compatibility on NeSI can matter. 

I would like to see if 5.4.5.3_new_gnu and 5.4.5.3 produce identical output on NeSI. Some reference to the nature of 5.4.5.3_new_gnu might be useful. Was there a code change or was it just built with a "new GCC" compiler (what version of GCC is *new*, by the way?).

In addition, as I raised while reviewing another PR of yours, I'd like to see more description for a PR. Took me some time and effort to follow what you were trying to achieve just by looking at the code change.
 
 
Screenshot from 2022-09-28 09-37-30

jasonmotha:
In person discussion about 5.4.5.3_new_gnu vs 5.4.5.3 -- key points:

- https://wiki.canterbury.ac.nz/display/QuakeCore/HF+Sim+Version
- https://wiki.canterbury.ac.nz/display/QuakeCore/HF+CUDA+Conversion
- Doesn't appear to be wiggle-for-wiggle comparable
- Viktor's new_gnu version is no longer being maintained.
- No publications used the new_gnu versions. Validation used Rob Graves versions; subduction used 6.0.3.x+ etc. Cybershake did use it for parts of the run. However, have kept the waveforms and have run a superceeded version since.

sungeunbae:
> In person discussion about 5.4.5.3_new_gnu vs 5.4.5.3 -- key points:
> 
> * https://wiki.canterbury.ac.nz/display/QuakeCore/HF+Sim+Version
> * https://wiki.canterbury.ac.nz/display/QuakeCore/HF+CUDA+Conversion
> * Doesn't appear to be wiggle-for-wiggle comparable for the same seed but equivalent statistically.
> * Viktor's new_gnu version is no longer being maintained.
> * No publications used the new_gnu versions. Validation used Rob Graves versions; subduction used 6.0.3.x+ etc. Cybershake did use it for parts of the run. However, have kept the waveforms and have run a superceeded version since.

Excellently summarised. No objections for replacing 5.4.5.3_new_gnu with 5.4.5.3.

sungeunbae (approved):

PR #455, opened 2022-10-02
Fixes to progress script

joelridden:
Ensures that the script will still display progress if files are missing so estimation can't be completed or other unexpected errors.
List is actually now an optional parameter. (When not specified grabs all faults from the DB)


workflow/automation/estimation/estimate_cybershake.py
```diff
@@ -283,88 +290,106 @@ def main(
...
+        nt = fault_sim_durations / dt
 
-    lf_ncores = (
-        np.ones(fault_names.shape[0], dtype=np.float32)
-        * platform_config[const.PLATFORM_CONFIG.LF_DEFAULT_NCORES.name]
-    )
+        lf_ncores = (
+            np.ones(fault_names.shape[0], dtype=np.float32)
+            * platform_config[const.PLATFORM_CONFIG.LF_DEFAULT_NCORES.name]
+        )
```
jpa198, line 302: These shouldn't raise FileNotFound, so shouldn't be in the try statement

workflow/automation/estimation/estimate_cybershake.py
```diff
@@ -283,88 +290,106 @@ def main(
...
+        lf_input_data = np.concatenate(
+            (
+                vm_params[:, :3],
+                nt.reshape(-1, 1),
+                fd_counts.reshape(-1, 1),
+                lf_ncores.reshape(-1, 1),
+            ),
+            axis=1,
+        )
```
jpa198, line 319: This should be in an 'else' statement, as there's no FNF error here either

workflow/automation/estimation/estimate_cybershake.py
```diff
@@ -283,88 +290,106 @@ def main(
...
+        hf_input_data = np.concatenate(
+            (
+                r_fd_counts[:, None],
+                r_nsub_stochs[:, None],
+                r_hf_nt[:, None],
+                r_hf_ncores[:, None],
+            ),
+            axis=1,
+        )
+    except UnboundLocalError or FileNotFoundError:
```
jpa198, line 354: Where is UnboundLocalError expected to come up?

jpa198 (dismissed):

sungeunbae (approved):

PR #465, opened 2023-01-13
Automated srf gen

jpa198:
#### Checklist


jpa198:
The one failing test depends on the parallel qcore PR

sungeunbae (commented):
Considering the significance and size of this PR, it deserves some description.

workflow/automation/install_scripts/install_realisation.py
```diff
@@ -0,0 +1,176 @@
...
+    parser.add_argument("--log_dir", type=Path)
+
+    vm_pert = parser.add_mutually_exclusive_group()
+    vm_pert.add_argument(
+        "--vm_perturbations",
+        action="store_true",
+        help="Use velocity model perturbations. If this is selected all realisations must have a perturbation file.",
+    )
+    vm_pert.add_argument(
+        "--ignore_vm_perturbations",
```
sungeunbae, line 20: Option name is slightly long, but I can accept this

sungeunbae: What happens both --vm_perturbations and --ignore_vm_perturnations are set? I can sense these two are not meant to be behave in opposite ways...or are they?

sungeunbae (commented):
Just have a comment in relation to --vm_perturbations argument. Other than that, can't fault anything. Well done.

workflow/automation/install_scripts/install_realisation.py
jpa198: On line 13 a mutually exclusive group is made, so argparse will prevent both options being chosen at the same time
The option name is inherited from the current install_cybershake.py script. But in this case, the script is not expected to be used by a user, only through the automated workflow so length of argument name is less important than clarity

joelridden (approved):

sungeunbae (approved):

PR #475, opened 2023-04-21
Don't scale core count for VM_pert

jpa198:
#### Checklist


joelridden (dismissed):

sungeunbae:
Give me some background please?

jpa198:
> Give me some background please?

With the current settings, this means that VM perturbation jobs can scale the number of cores requested for a job. VM perturbation maxes out at about 4-5 threads, so any core scaling would not have a positive impact on the run time (Actually negative as the WCT would be too short to finish and there's no checkponting). This change means that the number cores will not be scaled , meaning the scaled CH will be allocated to only WCT, instead of core count.

sungeunbae (dismissed):

joelridden (approved):

sungeunbae (approved):

PR #476, opened 2023-05-09
Update hf_sim.py

jpa198:
Correction based on reading HF source

#### Checklist

### DIFF
--- a/workflow/calculation/hf_sim.py
+++ b/workflow/calculation/hf_sim.py
@@ -160,7 +160,7 @@ def args_parser(cmd=None):
         help="dir containing site specific velocity models (1D). requires --site_specific",
     )
     # HF IN, line 14
-    arg("--vs-moho", help="depth to moho, < 0 for 999.9", type=float, default=999.9)
+    arg("--vs-moho", help="vs of moho layer, < 0 for 999.9", type=float, default=999.9)
     # HF IN, line 17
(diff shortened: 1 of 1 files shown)


claudio525 (approved):

sungeunbae (commented):
Do you mean we have used this argument incorrectly? Does it impact how we computed hf_sim?

jpa198:
> Do you mean we have used this argument incorrectly? Does it impact how we computed hf_sim?

We have never used this, it relies on the specified value being exceeded in the Vs (shear wave velocity) column of the 1d VM, I don't recall us specifying a lower value than the default in the past, so this functionality has never been used. The main purpose of this PR is to correct the help text for anyone trying to use it in the future. 
In the HF binary the last layer is used as the moho layer when no layer exceeds the specified value, so it's always defaulted to the last layer.

joelridden (approved):

sungeunbae:
> > Do you mean we have used this argument incorrectly? Does it impact how we computed hf_sim?
> 
> We have never used this, it relies on the specified value being exceeded in the Vs (shear wave velocity) column of the 1d VM, I don't recall us specifying a lower value than the default in the past, so this functionality has never been used. The main purpose of this PR is to correct the help text for anyone trying to use it in the future. In the HF binary the last layer is used as the moho layer when no layer exceeds the specified value, so it's always defaulted to the last layer.

Awesome. Happy to merge.

sungeunbae (approved):

PR #478, opened 2023-05-16
Only output timeseries at end of simulation

jpa198:
I've actually looked into how dump_itinc actually works, and it represents the intervals at which the timeseries information is saved to disk. This is not required for checkpointing, so only needs to be done at the end of the simulation. This will save significant time as we only need to write once.

#### Checklist


joelridden (dismissed):

joelridden (approved):

claudio525 (approved):

jasonmotha (approved):
very cool. And should stop a few issues with trying to run bb on incomplete simulations.

PR #480, opened 2023-05-30
Set database

jpa198:
#### Checklist


jpa198:
Is anyone interested in inheriting this script? 
The idea was that you could set the simulation database to a specific state after installation so that you could resume from pre-generated sources, or perform other steps without having to actually perform the prerequisites (LF2BB from pre-generated LF for example).

sungeunbae:
This has been in my wishlist for some time.  It's now mine!

joelridden:
I think it will be quite useful especially if we want to do re runs of data that we have semi saved etc

joelridden (approved):

PR #481, opened 2023-05-31
Amplification uncertainty

jpa198:
#### Checklist


workflow/calculation/bb_sim.py
```diff
@@ -461,6 +475,20 @@ def unfinished():
                     fmidbot=fmidbot,
                     version=site_amp_version,
                 )
+                if args.site_amp_uncertainty is not False:
+                    freqs = get_ft_freq(bb_dt, n2)
+                    if args.site_amp_uncertainty is None:
+                        lf_seed = hf_seed = None
```
joelridden, line 481: Would it be better to provide a flag for enabling the site amp uncertainty and then have another flag for the seed value to avoid the True/False identity with the None/int identity?

jpa198: So currently the behaviour is:
False: No uncertainty
True: uncertainty with random seed each time (not recorded)
integer: treated as True with a given seed (changed per realisation)

Adding a seed flag would cause the following behaviour:
False, no seed: no uncertainty
False, seed given: Raise exception?
True, no seed: Uncertainty with random seed
True, seed given: Uncertainty with given seed (Changed per rel)

My two issues are that a second flag increases the maintenance cost, and you now have to deal with the False, seed given case. I do think it would be clearer, but less concise to do it this way.
Ultimately I don't think it's a big deal either way.

joelridden (approved):

sungeunbae (approved):

PR #482, opened 2023-06-05
Update 18.5 configs to 5.4.5.3 HF

joelridden:
Changing version of hf to 5.4.5.3 in 18.5 older versions because of some issue when the hypocentre is too close to the boundary in the 1DVM.


joelridden:
> Will this change affect the reproducibility of old simulation result?

It would affect reproducability for cybershake runs that used this config, in the case of Robins simulation run 5.4.5.3 was actually used instead and so without this you would need to change the configs yourself to reproduce results.

I think it's still a good idea however to change these because it better aligns with results we are interested in reproducing (Robins past simulations) and it's more 'correct' as it's fixing an issue for the results.

If we really want to reproduce results for old simulations that isn't robins set then we can adjust as we need (better to do less adjusting for students using workflow imo)

sungeunbae:
> > Will this change affect the reproducibility of old simulation result?
> 
> It would affect reproducability for cybershake runs that used this config, in the case of Robins simulation run 5.4.5.3 was actually used instead and so without this you would need to change the configs yourself to reproduce results.
> 
> I think it's still a good idea however to change these because it better aligns with results we are interested in reproducing (Robins past simulations) and it's more 'correct' as it's fixing an issue for the results.
> 
> If we really want to reproduce results for old simulations that isn't robins set then we can adjust as we need (better to do less adjusting for students using workflow imo)

Cool - I think it is safe to assume it is @lee-robin 's result that is subject to the reproducibility.  Thanks for the clarification.

sungeunbae (approved):

lee-robin:
Fair point from Sung, I forgot about CyberShake when I suggested this change. I kept having to update the version for my validation runs since there was some odd behaviour with earlier versions (5.4.5.1 and 5.4.5.2) that only became apparent when we ran moderate Mw simulation validation (in 2020) but retrospectively we found it also affected some small Mw simulation validation too. This is why 5.4.5.3 first became the default version in 20.4.1 (https://wiki.canterbury.ac.nz/display/QuakeCore/Simulation).

One alternative is we can keep these the same and will just have to tell Ayushi to change her own one when she tries to reproduce my results. This might be less problematic in the unlikely case that we do want to reproduce something in the future. Keen to hear both of your thoughts on this?

sungeunbae:
> Fair point from Sung, I forgot about CyberShake when I suggested this change. I kept having to update the version for my validation runs since there was some odd behaviour with earlier versions (5.4.5.1 and 5.4.5.2) that only became apparent when we ran moderate Mw simulation validation (in 2020) but retrospectively we found it also affected some small Mw simulation validation too. This is why 5.4.5.3 first became the default version in 20.4.1 (https://wiki.canterbury.ac.nz/display/QuakeCore/Simulation).
> 
> One alternative is we can keep these the same and will just have to tell Ayushi to change her own one when she tries to reproduce my results. This might be less problematic in the unlikely case that we do want to reproduce something in the future. Keen to hear both of your thoughts on this?

Thanks @lee-robin for further clarification. I understand as long as your past simulation runs are concerned, we could associate hf 5.4.5.3 to gmsim ver 20.4.1, but we can't 100% rule out the possibility of somebody having run some simulations with hf 5.4.5.1 and 5.4.5.2 using gmsim ver 20.4.1 (and someone has to reproduce that result)? 

It is very unlikely, but if we want to be serious about the backward compatibility, we could create a new minor revision. For example, we keep 18.5.3.1 as it is, but a new minor revision 18.5.3.1.*1* with hf version 5.4.5.3 
ie.
```
18.5.3.1
hf version: 5.4.5.1
```
but
```
18.5.3.1.1
hf version: 5.4.5.3
```

We explicitly update our own records (eg. wiki etc) that Robin's simulation used this 18.5.3.1.1 revision and tell Ayushi to use this revision, not the original.

This way, In an unlikely event, if someone relying on the original version needs to rerun the calculation, that person won't be affected.

To be honest, I doubt this is a big issue, but we could use this opportunity to think of a future action for a similar situation.  
Your thoughts, @lee-robin ?

sungeunbae (approved):
Thanks @joelridden

sungeunbae:
@lee-robin  We have created new minor revisions to use hf version 5.4.5.3 while keeping the old versions using hf version 5.4.5.1 or 5.4.5.2 intact. That way we can avoid unforeseen backward compatibility issue, and advise Ayushi (or someone we can interact with) to use a version that we believe is good to reproduce the intended result.

claudio525 (approved):
Only have a rough idea whats going on here, but looks good

PR #483, opened 2023-07-24
Estimation fixes due to core hour scaling

joelridden:
Fixing few issues that popped up when core hours are being scaled.
Includes fix for run time adjustment after core count changes when in the green zone


claudio525 (approved):

sungeunbae (approved):
Excellent

PR #494, opened 2023-08-08
Skip loading sim_/fault_/vm_params when not ready

sungeunbae:
When a step in pre-processing gets executed (eg. srf_gen), some of *-params.yaml files are yet to be produced. But they are designed to be loaded in a cascade. and crash if there is one missing.

This fix will determine if the type of params.yaml should be loaded to perform a specified process type, and skip if not present.

### DIFF
--- a/workflow/automation/metadata/log_metadata.py
+++ b/workflow/automation/metadata/log_metadata.py
@@ -206,9 +206,20 @@ def main(args):
         )
         metadata_dict[const.MetadataField.run_time.value] = tdelta.total_seconds()
 
+    # Skip loading sim_params/fault_params/vm_params, as there are yet to exist
+    if args.proc_type in [
+        const.ProcessType.SRF_GEN.str_value,
+        const.ProcessType.INSTALL_REALISATION.str_value,
+        const.ProcessType.VM_PARAMS.str_value,
+        const.ProcessType.VM_GEN.str_value,
+        const.ProcessType.VM_PERT.str_value,
+        const.ProcessType.NO_VM_PERT.str_value,
+    ]:
+        return
     # Load the params
     params = utils.load_sim_params(
-        os.path.join(args.sim_dir, "sim_params.yaml"), load_vm=True
+        os.path.join(args.sim_dir, "sim_params.yaml"),
+        load_vm=True,
     )
 
     if "dt" in params["bb"]:
--- a/workflow/automation/metadata/log_metadata.py
+++ b/workflow/automation/metadata/log_metadata.py
@@ -207,7 +207,7 @@ def main(args):
         metadata_dict[const.MetadataField.run_time.value] = tdelta.total_seconds()
 
     # Skip loading sim_params/fault_params/vm_params, as there are yet to exist
-    if args.proc_type in [
+    if args.proc_type not in [
         const.ProcessType.SRF_GEN.str_value,
         const.ProcessType.INSTALL_REALISATION.str_value,
         const.ProcessType.VM_PARAMS.str_value,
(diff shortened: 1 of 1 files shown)


workflow/automation/metadata/log_metadata.py
```diff
@@ -206,9 +206,20 @@ def main(args):
...
+    # Skip loading sim_params/fault_params/vm_params, as there are yet to exist
+    if args.proc_type in [
+        const.ProcessType.SRF_GEN.str_value,
+        const.ProcessType.INSTALL_REALISATION.str_value,
+        const.ProcessType.VM_PARAMS.str_value,
+        const.ProcessType.VM_GEN.str_value,
+        const.ProcessType.VM_PERT.str_value,
+        const.ProcessType.NO_VM_PERT.str_value,
+    ]:
+        return
```
joelridden, line 218: Instead of just returning and not storing any of the previous metadata it could just skip all the metadata_dict entries that use the params. The store_metadata function at the bottom doesn't rely on any specific keys entered

sungeunbae: You mean we can move this block to the top of main() or even use this as a condition to call main() ?

sungeunbae: @joelridden  I struggle to understand your old comment ;) Help!

joelridden: Here is the entire main function with my explaination.
Essentially you are missing storing all the metadata that is done from lines 194 - 207 as the store_metadata function at the bottom is never called if you just return.
So Instead I just reversed the logic for a not and added all the params loading to the metadata if its not in that set of processes.

def main(args):
    # This should come from constants
    log_dir = os.path.join(args.sim_dir, "ch_log", const.METADATA_LOG_FILENAME)
    metadata_dict = getattr(args, METADATA_VALUES)
    # Determine run_time from start and end time
    if (
        const.MetadataField.start_time.value in metadata_dict.keys()
        and const.MetadataField.end_time.value in metadata_dict.keys()
    ):
        tdelta = datetime.strptime(
            metadata_dict[const.MetadataField.end_time.value],
            const.METADATA_TIMESTAMP_FMT,
        ) - datetime.strptime(
            metadata_dict[const.MetadataField.start_time.value],
            const.METADATA_TIMESTAMP_FMT,
        )
        metadata_dict[const.MetadataField.run_time.value] = tdelta.total_seconds()

    # Skip loading sim_params/fault_params/vm_params, as there are yet to exist
    if args.proc_type not in [
        const.ProcessType.SRF_GEN.str_value,
        const.ProcessType.INSTALL_REALISATION.str_value,
        const.ProcessType.VM_PARAMS.str_value,
        const.ProcessType.VM_GEN.str_value,
        const.ProcessType.VM_PERT.str_value,
        const.ProcessType.NO_VM_PERT.str_value,
    ]:
        # Load the params
        params = utils.load_sim_params(
            os.path.join(args.sim_dir, "sim_params.yaml"),
            load_vm=True,
        )
    
        if "dt" in params["bb"]:
            bb_dt = params["bb"]["dt"]
        else:
            bb_dt = min(params["dt"], params["hf"]["dt"])
        # params metadata for LF
        if args.proc_type == const.ProcessType.EMOD3D.str_value:
            metadata_dict[const.MetadataField.nt.value] = int(
                float(params["sim_duration"]) / float(params["dt"])
            )
            metadata_dict[const.MetadataField.nx.value] = params["nx"]
            metadata_dict[const.MetadataField.ny.value] = params["ny"]
            metadata_dict[const.MetadataField.nz.value] = params["nz"]
        # HF
        elif args.proc_type == const.ProcessType.HF.str_value:
            metadata_dict[const.MetadataField.nt.value] = int(
                float(params["sim_duration"]) / float(params["hf"]["dt"])
            )
            metadata_dict[const.MetadataField.nsub_stoch.value] = get_nsub_stoch(
                params["hf"]["slip"], get_area=False
            )
        # BB
        elif args.proc_type == const.ProcessType.BB.str_value:
            metadata_dict[const.MetadataField.dt.value] = bb_dt
        # IM_calc
        elif args.proc_type == const.ProcessType.IM_calculation.str_value:
            metadata_dict[const.MetadataField.nt.value] = int(
                float(params["sim_duration"]) / float(bb_dt)
            )
            # This should come from a constants file
            im_calc_csv_file = os.path.join(
                args.sim_dir, "IM_calc", "{}.csv".format(os.path.basename(args.sim_dir))
            )
            im_comp = list(pd.read_csv(im_calc_csv_file).component.unique().astype("U"))
            metadata_dict[const.MetadataField.im_comp.value] = im_comp
            metadata_dict[const.MetadataField.im_comp_count.value] = len(im_comp)
        # Advanced_IM
        elif args.proc_type == const.ProcessType.advanced_IM.str_value:
            metadata_dict[const.MetadataField.nt.value] = int(
                float(params["sim_duration"]) / float(bb_dt)
            )
    store_metadata(
        log_dir, args.proc_type, metadata_dict, sim_name=os.path.basename(args.sim_dir)
    )

sungeunbae: wow - you could have created a new PR yourself! haha. Thanks.

claudio525 (approved):

joelridden (approved):

PR #495, opened 2023-08-08
Use the system defined max wct if the estimated wallclock exceeds it

sungeunbae:
This change is probably not essential, but I had this bit added to the KISTI workflos codebass sometime. When a wallclock time is estimated, this gets compared with the system's maximum allowed wall clock time and goes with the less value.

### DIFF
--- a/workflow/automation/lib/shared_template.py
+++ b/workflow/automation/lib/shared_template.py
@@ -7,7 +7,7 @@ import qcore.constants as const
 from qcore.utils import load_sim_params
 from workflow.automation.lib.schedulers.scheduler_factory import Scheduler
 from workflow.automation.platform_config import platform_config
-
+from qcore import config
 
 def write_sl_script(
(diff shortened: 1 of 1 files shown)


joelridden (commented):
This shouldn't happen? the confine_wct_node_parameters function should correctly ensure it reduces the wall clock limit even with retries to be lower than the limit? Was this an issue before the changes were made to this function?

sungeunbae:
> This shouldn't happen? the confine_wct_node_parameters function should correctly ensure it reduces the wall clock limit even with retries to be lower than the limit? Was this an issue before the changes were made to this function?

Yes, this change was made before you recently looked at the related issue - I felt this change is no longer required, but just wanted to check with you.

sungeunbae:
I'm closing this one. I guess @joelridden 's earlier fix covers the case of concern.

PR #504, opened 2023-11-06
Update auto_submit.py

mrldupuis:
resolves error messages

#### Checklist

### DIFF
--- a/workflow/automation/execution_scripts/auto_submit.py
+++ b/workflow/automation/execution_scripts/auto_submit.py
@@ -51,7 +51,7 @@ def submit_task(
     # Metadata logging setup
     ch_log_dir = os.path.abspath(os.path.join(sim_dir, "ch_log"))
     if not os.path.isdir(ch_log_dir):
-        os.mkdir(ch_log_dir)
+        os.makedirs(ch_log_dir,exist_ok=True)
 
     runs_dir = sim_struct.get_runs_dir(root_folder)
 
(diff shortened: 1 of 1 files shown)


sungeunbae (dismissed):

sungeunbae (approved):

sungeunbae:
@claudio525 will you approve this asap?

joelridden (approved):
The logic of this does not make sense to me, the only way the previous result could be an error is if the sim directory did not exist which seems unlikely to ever be the case as it would need to make both the sim dir and the ch_log dir. The exist ok being True is odd too me too because it would only get into that if statement if it doesn't exist.

But that being said it won't break anything having this merged, so sure.

claudio525 (approved):

sungeunbae:
> The logic of this does not make sense to me, the only way the previous result could be an error is if the sim directory did not exist which seems unlikely to ever be the case as it would need to make both the sim dir and the ch_log dir. The exist ok being True is odd too me too because it would only get into that if statement if it doesn't exist.
> 
> But that being said it won't break anything having this merged, so sure.

Thanks - I think there was a corner case but can’t remember top of my head…

sungeunbae:
> The logic of this does not make sense to me, the only way the previous result could be an error is if the sim directory did not exist which seems unlikely to ever be the case as it would need to make both the sim dir and the ch_log dir. The exist ok being True is odd too me too because it would only get into that if statement if it doesn't exist.
> 
> But that being said it won't break anything having this merged, so sure.

Oh, I was reading this on phone and thought this is something funky I experienced previously and left a comment advocating this change, but I get the different view reading in on a big screen. I find it odd too. I thought this was to handle exceptions such as the presence of a symbolic link or a file of the same name, but it doesn't seem to be a case. I think if os.path.isdir() can go.

PR #507, opened 2023-12-05
Fix: Why --n_runs doesn't work as expected?

sungeunbae:
For various reasons, run_cybershake.py can crash or get interrupted by a user. I am running this command after a couple of reruns, so the DB and state is not clean.

```
python /nesi/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/execution_scripts/run_cybershake.py . task_config.yaml --n_runs 500
```
I want to keep 500 jobs in the queue, which is far below the maximum 1000 jobs accepted by NeSI.

However, I get

```
Exception in thread main auto submit:
Traceback (most recent call last):
  File "/opt/python/3.9.13.2/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/opt/python/3.9.13.2/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/scale_wlg_persistent/filesets/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/execution_scripts/auto_submit.py", line 586, in run_main_submit_loop
    submit_task(
  File "/scale_wlg_persistent/filesets/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/execution_scripts/auto_submit.py", line 98, in submit_task
    submit_lf_main(
  File "/scale_wlg_persistent/filesets/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/submit/submit_emod3d.py", line 105, in main
    submit_script_to_scheduler(
  File "/scale_wlg_persistent/filesets/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/lib/shared_automated_workflow.py", line 47, in submit_script_to_scheduler
    job_id = Scheduler.get_scheduler().submit_job(sim_dir, script, target_machine)
  File "/scale_wlg_persistent/filesets/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/lib/schedulers/slurm.py", line 119, in submit_job
    raise self.raise_exception(
workflow.automation.lib.schedulers.abstractscheduler.SchedulerException: An error occurred during job submission: sbatch: error: AssocMaxSubmitJobLimit
sbatch: error: Batch job submission failed: Job violates accounting/QOS policy (job submit limit, user's size and/or time limits)
```
, and run_cybershake.py exits.

There are indeed 1000 jobs currently in the queue.
```
(python3_maui) baes@maui01: /nesi/nobackup/nesi00213/RunFolder/sjn87/Canterbury/100mProductionRun$ squeue -u $USER |wc
   1000   11000  108000
```

Why is it getting AssocMaxSubmitJobLimit?  I thought we check the number of jobs in the queue and submit new jobs accordingly.

The reason turned out to be quite silly.  
Let's look at the run command again.

```
python $gmsim/workflow/workflow/automation/execution_scripts/run_cybershake.py . task_config.yaml --n_runs 500
```

Can you spot what is wrong here? 

```
(python3_maui) baes@maui01: /nesi/nobackup/nesi00213/RunFolder/sjn87/Canterbury/100mProductionRun$ python /nesi/project/nesi00213/Environments/sjn872033/workflow/workflow/automation/execution_scripts/run_cybershake.py --help
usage: run_cybershake.py [-h] [--sleep_time SLEEP_TIME] [--n_max_retries N_MAX_RETRIES] [--n_runs N_RUNS [N_RUNS ...]] [--log_folder LOG_FOLDER] [--debug] [--alert_url ALERT_URL] [--no-queue-monitor] root_folder user [config_file]

positional arguments:
  root_folder           The root directory of the simulation folder structure
  user                  Your username
  config_file           The location of the config file containing everything to be run
 ```
 
Gosh,  we only gave 2 positional arguments, which is still ok, as config_file is optional (it has ` nargs="?"` ). 

As a result, task_config.yaml becomes "user".  
 
This is an easy to make mistake, but we don't have an early detection. We never check if this is a valid user account name, and it just carries on.  This invalid "user" never bothers the system until it eventually arrives at slurm.py: check_queues() function. 

Here,  this "user" value is used to query the queue.  ie `squeue -u task_config.yaml`. Of course, the number of jobs in the queue under this user is 0, and it is always less than specified `--n_runs` value. Consequently we keep submitting more and more jobs until we hit AssocMaxSubmitJobLimit.

Proposed fix:

Option 1. Do nothing. This is a silly user mistake.  
Option 2.  Remove the "user" argument.
```
import getpass
user = getpass.getuser()
 ```
gives the user account.  I can't think of a situation where we execute run_cybershake.py with someone else's account, .so it makes sense to obtain the user account without relying on user's input
 
We will have to inform all users (eg. Sarah, Mike etc) to drop the second "user" argument from their recipebook for copy & paste execution.  This can be a little disruptive.
 
Option 3.  Check if the "user" argument is valid. Check this value belongs to the valid user list. , This, however, can be more complicated to implement without gaining much, especially we need to implement different method to obtain user list for different machine.

====================
After a quick team discussion, I have decided to go with Option2, and the fix has been pushed.

Once, merged, will update 

-  Cybershake manual : https://wiki.canterbury.ac.nz/pages/viewpage.action?pageId=90538503
-  researchers (Mike, Sarah in particular)

### DIFF
--- a/workflow/automation/execution_scripts/run_cybershake.py
+++ b/workflow/automation/execution_scripts/run_cybershake.py
@@ -1,5 +1,6 @@
 import argparse
 from datetime import datetime
+from getpass import getuser
 from logging import Logger, DEBUG
 from os.path import abspath, join
 import threading
@@ -204,7 +205,6 @@ def main():
     parser.add_argument(
         "root_folder", help="The root directory of the simulation folder structure"
     )
-    parser.add_argument("user", help="Your username")
     parser.add_argument(
         "config_file",
         help="The location of the config file containing everything to be run",
@@ -285,7 +285,7 @@ def main():
 
     scheduler_logger = qclogging.get_logger(name="scheduler", threaded=True)
     qclogging.add_general_file_handler(scheduler_logger, scheduler_log_file)
-    Scheduler.initialise_scheduler(user=args.user, logger=scheduler_logger)
+    Scheduler.initialise_scheduler(user=getuser(), logger=scheduler_logger)
 
     n_runs = 0
     if args.n_runs is not None:
--- a/workflow/automation/execution_scripts/run_cybershake.py
+++ b/workflow/automation/execution_scripts/run_cybershake.py
@@ -233,7 +233,7 @@ def main():
         nargs="+",
         help="The number of processes each machine can run at once. If a single value is given this is used for all "
         "machines, otherwise one value per machine must be given. The current order is: {}".format(
-            list(x.value for x in HPC)
+            list(x.name for x in HPC)
         ),
     )
     parser.add_argument(
--- a/workflow/automation/lib/schedulers/slurm.py
+++ b/workflow/automation/lib/schedulers/slurm.py
@@ -24,7 +24,6 @@ class Slurm(AbstractScheduler):
         if target_machine is None:
             target_machine = HPC[self.current_machine] #self.current_machine is str
 
-        print(f"################# {self.account} {target_machine} {self.current_machine} ###################")
         if isinstance(self.account, dict):
             account = self.account[target_machine.name]
         else:
@@ -115,9 +114,8 @@ class Slurm(AbstractScheduler):
                 raise e
             return jobid
         else:
-            cur_jobs = self.check_queues(True)
             raise self.raise_exception(
-                f"An error occurred during job submission: {err} Currently running {len(cur_jobs)} jobs"
+                f"An error occurred during job submission: {err}"
             )
 
     def cancel_job(self, job_id, target_machine=None):
--- a/workflow/automation/lib/schedulers/slurm.py
+++ b/workflow/automation/lib/schedulers/slurm.py
@@ -113,9 +113,13 @@ class Slurm(AbstractScheduler):
                 raise e
             return jobid
         else:
-            raise self.raise_exception(
-                f"An error occurred during job submission: {err}"
-            )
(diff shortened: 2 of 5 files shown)


joelridden:
Option 2, just for documentation sake

workflow/automation/execution_scripts/run_cybershake.py
```diff
@@ -233,7 +233,7 @@ def main():
         nargs="+",
         help="The number of processes each machine can run at once. If a single value is given this is used for all "
         "machines, otherwise one value per machine must be given. The current order is: {}".format(
-            list(x.value for x in HPC)
+            list(x.name for x in HPC)
```
sungeunbae, line 236: This change was to show ["mahuika", "maui"] instead of [1,2]

workflow/automation/lib/schedulers/slurm.py
```diff
@@ -22,7 +22,7 @@ def check_queues(self, user: bool = False, target_machine: HPC = None):
         )
 
         if target_machine is None:
-            target_machine = self.current_machine
+            target_machine = HPC[self.current_machine]  # self.current_machine is str
```
sungeunbae, line 25: self.current_machine is a string, whereas target_machine should be of HPC enum type to enable line 27
```
account = self.account[target_machine.name]
```

claudio525 (approved):

joelridden (approved):

PR #509, opened 2023-12-12
Empirical calculation re-enabled on NeSI Mahuika

sungeunbae:
Includes all the neccessary changes to re-enable empirical calculations. 

Tested with 

- Empirical Engine development branch: https://github.com/ucgmsim/Empirical_Engine/pull/108
- Qcore branch: https://github.com/ucgmsim/qcore/pull/294

Designed to be flexible. You can either supply pre-computed Z-values (auto-selected if it exists alongside .ll, .vs30 file in StationInfo directory. We plan to update Z-vlalues when we generate a new grid, or update basin models), or use on-the-fly estimated Z-values, the same behaviour as the old Empirical calculation. 
If it is a historical event (defined in root_params.yaml, historical: true), then it will be extracting event details from the NZ GMDB source file. Otherwise, it will use the srf info file sitting next to the .srf file.

In task_config.yaml, you need
```
Empirical: ALL
```

This PR also includes necessary environment changes. This calculation is only running on Mahuika, so Empirical_Engine installation is skipped on Maui. When Mahuika env is created, it will automatically clone and install oq-engine package on Mahuika env only, so we don't waste project storage for having two copies of oq-engine.

If Empirical calculation is to be added to the existing environment, follow the manual steps below.
1. Go to Mahuika terminal and navigate to $gmsim (environment root)
2. git clone Empirical_Engine repo and enter ../Empirical_Engine/
3. "pip install -r requirements.txt" then "pip install -e ./"


claudio525 (approved):

workflow/automation/submit/submit_empirical.py
```diff
@@ -22,16 +23,15 @@ def write_sl(sl_name, content):
...
-    run_data = [
-        (rel, fault)
-        for (rel, fault) in run_data
-        if rrup_file_exists(cybershake_folder, rel)
-    ]
+    run_data = [(rel, fault) for (rel, fault) in run_data]
+
     # determine NP
     # TODO: empirical are currently not parallel, update this when they are
     np = 1
```
joelridden, line 37: Just as a note this might need updating in the future once multiprocessing becomes a factor in empirical engine

joelridden (approved):

PR #511, opened 2024-05-09
Revised instructions on creating an environment

sungeunbae:
The steps described in this README is required to create a virtual environment on NeSI.
Found some incorrect stuffs while I walking through with Cesar and Andrew.

### DIFF
--- a/workflow/environments/README.md
+++ b/workflow/environments/README.md
@@ -20,7 +20,7 @@ Requirements: Setup github SSH keys for maui,
 
 A user specific environment of workflow, qcore, IMCalc, Empirical Engine and Pre-processing 
 can be created as follows:  
-1) Go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy
+1) Log in to Maui, and go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy. If you don't have it, `git clone` now.
 2) Navigate to ".../slurm_gm_workflow/workflow/environments/org/nesi"
 3) Load the correct version of Python module. (** IMPORTANT **). For example, Python 3.9, try the following.
    ```bash
--- a/workflow/environments/README.md
+++ b/workflow/environments/README.md
@@ -24,8 +24,8 @@ can be created as follows:
 ```bash
 git clone git@github.com:ucgmsim/slurm_gm_workflow.git
 ```
-2) Navigate to ".../slurm_gm_workflow/workflow/environments/org/nesi"
-3) Load the correct version of Python module. (** IMPORTANT **). For example, Python 3.9, try the following.
+2) Navigate to the copy of your workflow "slurm_gm_workflow/workflow/environments/org/nesi"
+3) Load the correct version of Python module. (**IMPORTANT**). For example, Python 3.9, try the following.
    ```bash
    module load cray-python cray-hdf5-parallel/1.12.2.3
    ```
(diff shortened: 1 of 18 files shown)


claudio525 (approved):

workflow/environments/README.md
```diff
@@ -20,7 +20,7 @@ Requirements: Setup github SSH keys for maui,
 
 A user specific environment of workflow, qcore, IMCalc, Empirical Engine and Pre-processing 
 can be created as follows:  
-1) Go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy
+1) Log in to Maui, and go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy. If you don't have it, `git clone` now.
 2) Navigate to ".../slurm_gm_workflow/workflow/environments/org/nesi"
```
lispandfound, line 27: This `...` is relative, but you don't say relative to where. In my case I:

1. Cloned `slurm_gm_workflow` in my home directory
2. Navigated to `slurm_gm_workflow/workflow/environments/org/nesi`

I think this should not have a relative path. Maybe explicitly state to work from `$HOME` and then navigate to 
`~/slurm_gm_workflow/workflow/environments/org/nesi`

workflow/environments/README.md
```diff
@@ -20,13 +20,16 @@ Requirements: Setup github SSH keys for maui,
 
 A user specific environment of workflow, qcore, IMCalc, Empirical Engine and Pre-processing 
 can be created as follows:  
-1) Go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy
+1) Log in to Maui, and go to an existing copy of workflow (eg. can be your home) - it is a good idea to update this copy. If you don't have it, 
+```bash
+git clone git@github.com:ucgmsim/slurm_gm_workflow.git
+```
 2) Navigate to ".../slurm_gm_workflow/workflow/environments/org/nesi"
 3) Load the correct version of Python module. (** IMPORTANT **). For example, Python 3.9, try the following.
```
lispandfound, line 28: NB: You should use `**IMPORTANT**` not `** IMPORTANT **` to bold the text. 

**IMPORTANT** vs ** IMPORTANT **

lispandfound: If bold is actually what you want, that is.

lispandfound (changes requested):

joelridden (approved):
Good changes, what Jake said about some other sections also could be good to change as well. But approve the rest.

Also will need the Black upgrade run over this code, hasnt been changed in a while

sungeunbae:
All Python changes are unrelated to the update to the README - just to keep Black happy.

joelridden (approved):

claudio525 (approved):

lispandfound (approved):

sungeunbae:
@AndrewRidden-Harper  You have followed this README to set up the virtual enviornment on NeSI. I made a few changes for the issues I noticed while going through this page.

workflow/environments/README.md
AndrewRidden-Harper: I agree that changing this to include the full path might be better to avoid potential confusion

AndrewRidden-Harper (approved):
Looks good. My only comment is that I agree with Jake's comment that using the full path might be better to avoid potential confusion

PR #512, opened 2024-05-21
Fixing "Database is locked" issue

sungeunbae:
@sarahjaneneill has been witnessing `database is locked` sqlite3.OperationalError frequently.

```
2024-02-14 02:03:11,935 - queue monitor - squeue tasks: 42576994 PD, 42576995 PD, 42576996 PD,...line 417, in get_runnable_tasks
    entries = cur.execute(
sqlite3.OperationalError: database is locked
2024-02-14 02:03:27,323 - MainThread - The main auto_submit thread has terminated, and all...
```
and this is my attempt to address the issue.

Reference: https://stackoverflow.com/questions/3172929/operationalerror-database-is-locked

I changed the code to release the connection as soon as one transaction is completed. Also gave 10 secs timeout before screaming "database is locked".

Tested by Sarah, and myself - the issue is not happening while burning half million core hours.

### DIFF
--- a/workflow/automation/execution_scripts/query_mgmt_db.py
+++ b/workflow/automation/execution_scripts/query_mgmt_db.py
@@ -5,6 +5,7 @@ A script that queries a slurm mgmt db and returns the status of a task
 """
 
 import argparse
+import os
 from typing import Union, List
 
 from workflow.automation.lib import MgmtDB
@@ -47,17 +48,27 @@ def state_table_query_builder(
     ordering: Union[bool, str, List[str]] = False,
 ):
     """
-    Creates queries for the state table.
-    For data sanitation reasons variables are passed in separately to the query string (No Bobby tables here)
-    :param what: The columns to be selected
-    :param state: Either true for one, or the number of states to be selected
-    :param process_type: Either true for one, or the number of states to be selected
-    :param run_name_exact: If an exact pattern match is required
-    :param run_name_similar: If a pattern match is required, to be used with '%' as the sqlite wildcard character
-    :param run_name_disimilar: If a pattern non-match is required, to be used with '%' as the sqlite wildcard character
-    :param task_id: Either true for one, or the number of states to be selected
-    :param ordering: The columns to order the results by
-    :return: The string to be used as the query
+    Creates queries for the state table. For data sanitation reasons variables are passed in separately to the query string.
+
+    Parameters
+    ----------
+    what : Union[str, List[str]]
+        The columns to be selected.
+    state : Union[bool, int], optional
+        Either true for one, or the number of states to be selected, by default False.
+    process_type : Union[bool, int], optional
+        Either true for one, or the number of states to be selected, by default False.
+    run_name_type : MgmtDB.ComparisonOperator, optional
+        If an exact pattern match is required, by default None.
+    task_id : Union[bool, int], optional
+        Either true for one, or the number of states to be selected, by default False.
+    ordering : Union[bool, str, List[str]], optional
+        The columns to order the results by, by default False.
+
+    Returns
+    -------
+    str
+        The string to be used as the query.
     """
     if isinstance(what, str):
         what = [what]
@@ -415,10 +426,9 @@ def main():
     if args.mode_help:
         print_mode_help()
         exit()
-    f = args.run_folder
+
     run_name = args.run_name
     mode = args.mode
-    db = MgmtDB.connect_db(f)
 
     query_mode = QueryModes()
     if "error" in mode:
@@ -433,7 +443,11 @@ def main():
     if "retry_max" in mode:
         query_mode.retry_max = True
 
-    print_run_status(db, run_name, query_mode, args.config)
+    with MgmtDB.connect_db_ctx(
+        os.path.join(args.run_folder, "slurm_mgmt.db"),
+        pragmas=["synchronous = EXTRA", "integrity_check"],
+    ) as cur:
+        print_run_status(cur, run_name, query_mode, args.config)
 
 
 if __name__ == "__main__":
--- a/workflow/automation/execution_scripts/query_mgmt_db.py
+++ b/workflow/automation/execution_scripts/query_mgmt_db.py
@@ -5,7 +5,7 @@ A script that queries a slurm mgmt db and returns the status of a task
 """
 
 import argparse
-import os
+from pathlib import Path
 from typing import Union, List
 
 from workflow.automation.lib import MgmtDB
@@ -444,7 +444,7 @@ def main():
         query_mode.retry_max = True
 
     with MgmtDB.connect_db_ctx(
-        os.path.join(args.run_folder, "slurm_mgmt.db"),
+        Path(args.run_folder) / "slurm_mgmt.db",
         pragmas=["synchronous = EXTRA", "integrity_check"],
     ) as cur:
         print_run_status(cur, run_name, query_mode, args.config)
--- a/workflow/automation/lib/MgmtDB.py
+++ b/workflow/automation/lib/MgmtDB.py
@@ -439,7 +439,7 @@ class MgmtDB:
         while len(runnable_tasks) < task_limit and offset < entries:
             with connect_db_ctx(self._db_file) as cur:
                 db_tasks = cur.execute(
-                    """SELECT run_name, proc_type 
+                    """SELECT proc_type, run_name 
                               FROM status_enum, state 
                               WHERE state.status = status_enum.id
                                AND proc_type IN (?{})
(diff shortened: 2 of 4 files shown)


claudio525 (dismissed):

sungeunbae:
@claudio525 haha - you are overzealous like SQLite locking mechanism ;) Still in draft stag, haha

workflow/automation/lib/MgmtDB.py
```diff
@@ -53,7 +53,11 @@ def connect_db_ctx(db_file, verbose=False):
 
     A commit is run at the end of the context.
     """
-    conn = sql.connect(db_file)
+    # timeout parameter specifies how long the connection should wait for the lock to
+    # go away until raising an exception. Default is 5 secs
+    # https://stackoverflow.com/a/8618328/2005856
+    # https://docs.python.org/2/library/sqlite3.html#sqlite3.connect
+    conn = sql.connect(db_file, timeout=10)
```
lispandfound, line 60: I don't like the `import sqlite3 as sql`. I'd rather see `sqlite3.connect` and at least know it came from the sqlite3 module. When I saw this I thought another library might be involved.

lispandfound: Also, what happens here if the connection is opened but then the rest of this function fails? In this case a connection to the database (and potentially a lock) would be leaked. Don't know if it affects your DB lock problems, but maybe in the interest of cleaning up all loose ends this connection should be freed in the case of an error with the rest of the function.

lispandfound: Oh ok nevermind I can't read that is accounted for.

workflow/automation/lib/MgmtDB.py
```diff
@@ -351,14 +355,14 @@ def add_retries(self, n_max_retries: int):
...
-        with connect_db_ctx(self._db_file) as cur:
-            for key, fail_count in failure_count.items():
-                if any([x >= n_max_retries for x in fail_count.values()]):
-                    continue
-                run_name, proc_type = key.split("__")
-                # Gets the number of entries for the task with state in [created, queued, running or completed]
-                # Where completed has enum index 5, and the other 4 less than this
-                # If any are found then don't add another entry
+        for key, fail_count in failure_count.items():
+            if any([x >= n_max_retries for x in fail_count.values()]):
```
lispandfound, line 359: Could rewrite as `max(fail_count.values()) >= x`

workflow/automation/lib/MgmtDB.py
```diff
@@ -351,14 +355,14 @@ def add_retries(self, n_max_retries: int):
...
+                continue
+            run_name, proc_type = key.split("__")
+            # Gets the number of entries for the task with state in [created, queued, running or completed]
+            # Where completed has enum index 5, and the other 4 less than this
+            # If any are found then don't add another entry
```
lispandfound, line 364: I find explanation confusing. What does enum mean? Why must the other four (three? created, queued, running?) have index less than 5?

workflow/automation/lib/MgmtDB.py
```diff
@@ -367,8 +371,8 @@ def add_retries(self, n_max_retries: int):
                     "AND status <= (SELECT id FROM status_enum WHERE state = 'completed') ",
                     (run_name, proc_type),
```
lispandfound, line 372: Could be made more efficient (and clearer) with a SQL JOIN:
```sql
SELECT COUNT(*)
FROM state s
JOIN status_enum se ON s.status <= se.id
WHERE se.state = 'completed';
```

lispandfound: double check that does the same thing (I ran it in the SQL DB engine of "my head"), but I think it does.

workflow/automation/lib/MgmtDB.py
```diff
@@ -441,15 +446,15 @@ def get_runnable_tasks(
...
-                        and "{}__{}".format(*task) not in tasks_waiting_for_updates
-                    ]
-                )
-                offset += 100
+            runnable_tasks.extend(
+                [
+                    (*task, self.get_retries(*task, get_WCT=True))
+                    for task in db_tasks
+                    if self._check_dependancy_met(task, logger)
+                    and "{}__{}".format(*task) not in tasks_waiting_for_updates
```
lispandfound, line 454: Rather than `'{}__{}'.format(*task)` could just have `'__'.join(task)`

lispandfound: Or, alternatively, be a bit more explicit in this loop and unwrap `task` in the for loop.

lispandfound (changes requested):
Could you also delete the `enum_to_list` function that is not used in this module, and clean up the unused imports?

workflow/automation/lib/MgmtDB.py
sungeunbae: can be simply `status <= 5`.

workflow/automation/lib/MgmtDB.py
sungeunbae: https://github.com/ucgmsim/qcore/blob/f57aae21c1489315369c9791a851c991ffb06c4c/qcore/constants.py#L448
Screenshot 2024-05-31 at 2 45 49 PM

sungeunbae (commented):
I'm a bit reluctant to touch to make too many changes here. When we introduce Pegasus, all these stuffs will be taken out.

workflow/automation/lib/MgmtDB.py
lispandfound: oh I didn't see that wasn't your code. Ignore this then.

workflow/automation/lib/MgmtDB.py
lispandfound: ok @sungeunbae if you don't want to change that because of pegasus that's fine. Maybe link to the latest sqlite3 docs rather than Python 2 docs? https://docs.python.org/3/library/sqlite3.html#sqlite3.connect

sungeunbae: Good suggestion!

joelridden (dismissed):

AndrewRidden-Harper (approved):

sungeunbae:
Please wait until I add some more commits

sungeunbae:
@claudio525 @joelridden @lispandfound @AndrewRidden-Harper 

Your opinion please.

This PR contains two parts of work. 
Part 1. Initially I was focusing on releasing cursor as soon as each query is done. I identified particular pathway Sarah was running into this issue, and unbundled heavy queries, but other heavy queries were left unchanged. Despite this effort, during the early stage of her "sprint", Sarah was still witnessing the issue, perhaps not as frequently as before.

Then I discovered Part 2, db.connect(...timeout=10) solution. 

Since Part 2 was added, we managed to burn 0.5 million core hours without witnessing the issue occurring. 

Now I have a reason to believe Part 1 was probably not necessary at all. 

DB handling is a high-risk area and with Pegasus-integration on horizon, I'm inclined to minimize the extent of change. So I'm proposing to close this PR without merge, and address this issue in a separate branch (https://github.com/ucgmsim/slurm_gm_workflow/tree/db_timeout_increase) that contains just the Part 2.  

I'll be in close contact with Sarah and Mike and see if the issue remains/re-occurs, and handle this case accordingly.

joelridden:
> @claudio525 @joelridden @lispandfound @AndrewRidden-Harper
> 
> Your opinion please.
> 
> This PR contains two parts of work. Part 1. Initially I was focusing on releasing cursor as soon as each query is done. I identified particular pathway Sarah was running into this issue, and unbundled heavy queries, but other heavy queries were left unchanged. Despite this effort, during the early stage of her "sprint", Sarah was still witnessing the issue, perhaps not as frequently as before.
> 
> Then I discovered Part 2, db.connect(...timeout=10) solution.
> 
> Since Part 2 was added, we managed to burn 0.5 million core hours without witnessing the issue occurring.
> 
> Now I have a reason to believe Part 1 was probably not necessary at all.
> 
> DB handling is a high-risk area and with Pegasus-integration on horizon, I'm inclined to minimize the extent of change. So I'm proposing to close this PR without merge, and address this issue in a separate branch (https://github.com/ucgmsim/slurm_gm_workflow/tree/db_timeout_increase) that contains just the Part 2.
> 
> I'll be in close contact with Sarah and Mike and see if the issue remains/re-occurs, and handle this case accordingly.

Could Part 1 still help though? I don't see any reason why we should not make those changes to only have the cursor when we need it. Better to just have these changes where we know it performed well rather than just assume it was just Part 2and ignore Part 1 but maybe without it still run into issues even with the timeout. Sure pegasus is coming but it's not a major complex change that will make anything more difficult at all.

sungeunbae:
I have decided to go ahead with this this PR. The combined effect of Part I and Part 2 is positive, which is "good" enough for now.
Discovered a bug while making some cosmetic changes following Jake's feedback, which has been addressed - I don't know how/why it worked in the past. Tested on Maui after the bug fix briefly, and it seems to be working ok.

claudio525 (dismissed):

lispandfound (dismissed):

workflow/automation/execution_scripts/query_mgmt_db.py
```diff
@@ -433,7 +443,11 @@ def main():
     if "retry_max" in mode:
         query_mode.retry_max = True
 
-    print_run_status(db, run_name, query_mode, args.config)
+    with MgmtDB.connect_db_ctx(
+        os.path.join(args.run_folder, "slurm_mgmt.db"),
```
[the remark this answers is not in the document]
sungeunbae: Haha, yes, I considered this, but I didn't proceed because there were a bit too many to update. Sure, why not. Pushing some changes now.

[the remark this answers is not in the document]
sungeunbae: @joelridden

joelridden (approved):

claudio525 (approved):

lispandfound (approved):

sungeunbae:
@AndrewRidden-Harper  ? will you have a look at this one? It is perhaps not the area you have much experience with, but just to give you some idea about

PR #514, opened 2024-06-05
Move load_sim_params to workflow

lispandfound:
Required for ucgmsim/qcore#309. Moves the `load_sim_params` function to workflow so that we can remove it from qcore.


claudio525 (approved):

sungeunbae (approved):

joelridden (commented):
Only thing I'd say is to maybe have the sim_params.py in the automation folder of workflow

workflow/sim_params.py
```diff
@@ -0,0 +1,99 @@
...
+    load_fault: Union[bool, str] = True,
+    load_root: Union[bool, str] = True,
+    load_vm: Union[bool, str] = True,
+) -> dict:
+    """Load all necessary parameters for a simulation.
+
+    Parameters
+    ----------
+    sim_yaml_path : Union[bool, str]
+        Path to sim_params.yaml or a falsy value to not load it.
```
AndrewRidden-Harper, line 56: I think this should be falsy --> False ?

workflow/sim_params.py
```diff
@@ -0,0 +1,99 @@
...
+    load_fault : Union[bool, str]
+        Either True, the path to fault_params.yaml or a falsy value to not load it.
```
AndrewRidden-Harper, line 58: falsy --> false ?

AndrewRidden-Harper (approved):

workflow/sim_params.py
lispandfound: No that's a [well-defined term](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false).

workflow/automation/submit/submit_sim_imcalc.py
```diff
@@ -38,7 +35,7 @@ def submit_im_calc_slurm(
     two dictionaries, the passed in one has higher priority.
     """
     # Load the yaml params
-    params = utils.load_sim_params(
+    params = sim_params.load_sim_params(
```
joelridden, line 38: missing import

joelridden (approved):

joelridden (approved):

AndrewRidden-Harper (approved):

claudio525 (approved):

sungeunbae (approved):

PR #520, opened 2024-06-26
correctly utilise simulation root argument to create3d

### LINKED ISSUE #518: create_e3d.py does not load sim_params correctly
It assumes it is called with sim params in the current directory, rather than using qcore to load the sim params file. This can create bugs if the assumption is incorrect (i.e. it is run from outside the `Runs` folder).

lispandfound:
Fixes #518

### DIFF
--- a/workflow/calculation/create_e3d.py
+++ b/workflow/calculation/create_e3d.py
@@ -8,11 +8,9 @@ USAGE: edit params.py (not this file), execute this file from the same directory
 ISSUES: remove default values in e3d_default.par where not needed.
 """
 
-from __future__ import print_function
 
 import argparse
 import os.path
-import sys
 from logging import Logger
 from os.path import basename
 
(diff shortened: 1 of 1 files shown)


AndrewRidden-Harper (approved):

sungeunbae (changes requested):
I understand create_e3d.py is always used by importing it instead of running it as a script, so this change and introducing SIM_ROOT parameter doesn't cause any degradation.

Can you remove sys.path.append() line (line 28)? Then could os.path could be removed.

lispandfound:
> I understand create_e3d.py is always used by importing it instead of running it as a script, so this change and introducing SIM_ROOT parameter doesn't cause any degradation.
> 
> Can you remove sys.path.append() line (line 28)? Then could os.path could be removed.

Done. Also removed a `from __future__ import print_function` that was not required anymore.

claudio525 (approved):

joelridden (approved):

sungeunbae (approved):

PR #521, opened 2024-07-02
Collect estimated time and time actually spent

sungeunbae:
I want to add this script to estimation tool kit in the workflow. The current wall-clock estimation is not very accurate, and makes it difficult to estimate the total core hours to run a Cybershake-styled large set of simulations. We also don't collect actual CPU time spent, and losing the crucial information of resource usage as well as the opportunity to improve the estimation quality.

This script will interact with SLURM (using sacct command, `-n` removes the header), 
```
$ sacct -j 4088784,4088640 --format="JobID,Elapsed,TimeLimit,AllocCPUS" -n
4088640        02:10:42   08:17:00        160
4088640.bat+   02:10:42                    80
4088640.ext+   02:10:42                   160
4088640.0      02:10:06                   160
4088784        00:03:38   00:30:00         80
4088784.bat+   00:03:38                    80
4088784.ext+   00:03:38                    80
 4088784.0      00:03:19                    80
```
This command gives more than we need - after we remove lines for steps of a job (.batch, .external, .0), we have
```
4088640        02:10:42   08:17:00        160
4088784        00:03:38   00:30:00         80
```
Here, 02:10:42 and 00:03:38 correspond to the CPU time used.
Wall clock time requested were 08:17:00 and 00:30:00 respectively. 160 and 80 mean the number of CPUs.

This script will consult SQLite3 DB, and collect all "completed" jobs, then query with "sacct" command to collect the estimated/actual resource usage, then create a CSV file.

Example run:
```
(python3_maui) baes@maui02: /nesi/nobackup/nesi00213/RunFolder/Cybershake/v24p6$ python $gmsim/workflow/workflow/automation/estimation/get_est_vs_used_cputime.py . list_crustal.txt 
Record found in DB : 4302 entries
CSV file est_vs_used_cpu_time.csv created successfully!
Total CPU hours needed for all realisations on maui : 3386491.20 hours
Total CPU hours needed for all realisations on mahuika : 5386.97 hours
```

### DIFF
--- a/workflow/automation/estimation/get_est_vs_used_cputime.py
+++ b/workflow/automation/estimation/get_est_vs_used_cputime.py
@@ -153,7 +153,7 @@ def main():
         assert job_id is not None, f"{run_name} {proc_type_str}"
         # Execute sacct command
         for machine in ["maui", "mahuika"]:
-            cmd = f'sacct -j {job_id} -M {machine} --format="JobName,time_used,time_requested,AllocCPUS"'
+            cmd = f'sacct -j {job_id} -M {machine} --format="JobName,Elapsed,TimeLimit,AllocCPUS"'
             sacct_output = subprocess.check_output(
                 cmd,
                 shell=True,
--- a/workflow/automation/estimation/get_est_vs_used_cputime.py
+++ b/workflow/automation/estimation/get_est_vs_used_cputime.py
@@ -4,7 +4,7 @@ Collect estimated vs used CPU time for Cybershake-styled runs
 This is particularly useful to accurately estimate the core hours needed for the entire set of Cybershake-styled
 simulations. One can run the MEDIAN simulations only, and then run this script to collect the estimated core hours and
 the cpu time actually used.
-By multiplying the time_used by the number of realisations that is (optionally) supplied with --list argument,
+By multiplying the time_used by the number of realisations,
 we can accurately estimate the total core hours required to run the entire set of simulations.
 
 This script is also useful to assess the quality of wall clock time estimation. If the estimated time is significantly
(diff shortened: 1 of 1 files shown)


AndrewRidden-Harper (dismissed):
looks good

lispandfound (changes requested):
I understand this is just an impromptu script and not a core part of the workflow, so I have only made small comments/suggestions.

joelridden (commented):
I think just taking a quick look we already have something for this
https://github.com/ucgmsim/slurm_gm_workflow/blob/master/workflow/automation/metadata/collect_metadata.py
But I don't think it has all of the data that's collected here, might be good to merge though instead of having 2 different metadata collection scripts which could cause some confusion. The other collect metadata script does gain extra metadata that can be used to re-estimate estimation better than what we just have here.

As a reminder the 30 minutes is also a min allowed wall clock time to set

sungeunbae:
> I think just taking a quick look we already have something for this https://github.com/ucgmsim/slurm_gm_workflow/blob/master/workflow/automation/metadata/collect_metadata.py But I don't think it has all of the data that's collected here, might be good to merge though instead of having 2 different metadata collection scripts which could cause some confusion. The other collect metadata script does gain extra metadata that can be used to re-estimate estimation better than what we just have here.
> 
> As a reminder the 30 minutes is also a min allowed wall clock time to set

Good point. There were a few core differences, which prompted me to write a new script, instead of fixing the existing solution. (eg. when I have no sim_params.yaml for all relisations, it will break) but I can see your rationale.

A few key differences include
1. my script allows to run median event only, and extrapolate the estimation for the entire rels.
2. it directly interact with SLURM, instead of relying on slurm_mgmt.db. If some jobs were managed outside the run_cybershake due to some issues and needed manual intervention, the DB may not necessarily have the best up-to-date data. 

I agree there are some commonalities between two, and it may seem logical to merge or let them share some code. I'm happy to withdraw this PR, but let's keep this branch (for future reference). I don't want to lose some SLURM techniques I learned while writing this code :)

PR #526, opened 2024-09-05
Skip e3d.par when moving tmp files

sungeunbae:
When EMOD3D fails, the slurm/PBS script relocate incomplete output files to a temporary directory for later debug.
However, it also moves e3d.par, which will make the restart attempt futile as it will fail instantly if e3d.par is missing.

This change hopefully fixes this issue

### DIFF
--- a/workflow/automation/templates/run_emod3d.sl.template
+++ b/workflow/automation/templates/run_emod3d.sl.template
@@ -52,5 +52,8 @@ else
     echo "Completion test failed, moving all files to $backup_directory"
     echo "Failure reason: $res"
     mkdir -p $backup_directory
-    mv {{sim_dir}}/LF/* $backup_directory
+    shopt -s extglob #enable !() option used below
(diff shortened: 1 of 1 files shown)


joelridden (commented):
Is this able to be tested to "fix" this issue? Just to be sure there wouldn't need to be a 2nd PR for a fix

sungeunbae:
> Is this able to be tested to "fix" this issue? Just to be sure there wouldn't need to be a 2nd PR for a fix

Sure - I will force it to fail and test this fix.

sungeunbae:
> > Is this able to be tested to "fix" this issue? Just to be sure there wouldn't need to be a 2nd PR for a fix
> 
> Sure - I will force it to fail and test this fix.

@joelridden 
On NeSI, `/nesi/nobackup/nesi00213/RunFolder/Cybershake/v24p9/Runs/HikWgtnmax/HikWgtnmax`, the EMOD3D job failed due to out of memory, but e3d.par file stayed - the fix works as intended.

joelridden (approved):

PR #533, opened 2024-10-15
Utilize SRF to call validate_vm.py

sungeunbae:
During vm_gen step, the .sl/.pbs scripts already know where SRF is, but it is not utilized to call validate_vm.py, and as a result, we don't check against the fault geometry. 
This PR fixes the problem.


joelridden (approved):

lispandfound (approved):

sungeunbae:
@claudio525  Any comments?

claudio525 (approved):

PR #540, opened 2026-03-04
Ensure vs30 values are written to disk and verified before exit

AndrewRidden-Harper:
This adds to @sungeunbae 's [recent HF fixes ](https://github.com/ucgmsim/slurm_gm_workflow/pull/539)
We were also finding that Vs30 values were sometimes missing from the output file, causing the BB stage to crash with a math error. The root cause seems to be that vs30 values were not completely written to the file before the program finished. To fix this, I've added flush and fsync, which should ensure that the writes have completed before the program finishes. I also added some validation checks for missing vs30 values at the end.

### DIFF
--- a/workflow/calculation/hf_sim.py
+++ b/workflow/calculation/hf_sim.py
@@ -513,6 +513,9 @@ if __name__ == "__main__":
                 out.seek(HEAD_STAT - 2 * FLOAT_SIZE, 1)
                 e_dist[i].tofile(out)
                 vs.tofile(out)
+            # Ensure vs30 values are fully written to disk before MPI terminates the process
+            out.flush()
+            os.fsync(out.fileno())
 
     # distribute work in a round-robin fashion across ranks for optimisation
     # if size=4, rank 0 takes [0,4,8...], rank 1 takes [1,5,9...], rank 2 takes [2,6,10...],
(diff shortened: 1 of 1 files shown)


workflow/calculation/hf_sim.py
```diff
@@ -513,6 +513,9 @@ def run_hf(
                 out.seek(HEAD_STAT - 2 * FLOAT_SIZE, 1)
                 e_dist[i].tofile(out)
                 vs.tofile(out)
+            # Ensure vs30 values are fully written to disk before MPI terminates the process
```
sungeunbae, line 516: vs30 values?

sungeunbae (commented):
I am a little confused here - do you mean HF.bin output file?

workflow/calculation/hf_sim.py
AndrewRidden-Harper: @sungeunbae Sorry, I meant vs values

AndrewRidden-Harper:
I just discovered a problem with the extra validation check I put in hf_sim.py. I've converted this back to draft until my fix is confirmed, then I will include it and mark is as ready for review again

AndrewRidden-Harper:
This PR actually doesn't reliably solve the problem of vs values not being written to the HF.bin file on RCH and we actually don't need it to work on RCH because we can just run it on NeSI where it always works properly, so I'll close this PR

